PR #13, opened 2014-08-11
[MIG] Translated changes from 7.0 branch

pedrobaeza:
After the PEP8 fix. Also a new module for 7.0 lands in `__unported__` dir.

### DIFF
--- a/__unported__/document_page_multi_company/__openerp__.py
+++ b/__unported__/document_page_multi_company/__openerp__.py
@@ -40,4 +40,4 @@ This module adds a company field to document page and the multi-company rule.
     'auto_install': False,
     'images': [],
 }
-# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
+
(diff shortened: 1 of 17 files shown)


hbrunn:
ah, I think given the modules live in **unported**, it doesn't hurt, but could you set installable: False for good order?

pedrobaeza:
You're right. Done.

hbrunn:
:+1:

__unported__/document_page_multi_company/__openerp__.py
```diff
@@ -40,3 +40,4 @@
     'auto_install': False,
     'images': [],
 }
+# vim:expandtab:smartindent:tabstop=4:softtabstop=4:shiftwidth=4:
```
bwrsandman: Don't need this addition

pedrobaeza: This is because I have copied as is from 7.0. In 7.0 must be this line. I'm removing it.

pedrobaeza:
I have removed comments for all files.

gurneyalex:
:+1:

bwrsandman:
:+1:

PR #15, opened 2014-09-04
[ADD] attachment_preview

hbrunn:


attachment_preview/__openerp__.py
```diff
@@ -0,0 +1,60 @@
...
+#
+##############################################################################
+{
+    "name": "Preview attachments",
+    "version": "1.0",
+    "author": "Therp BV",
+    "license": "AGPL-3",
+    "complexity": "normal",
+    "description": """
+This addons allows to preview attachments supported by http://viewerjs.org.
```
pedrobaeza: s/addons/addon. It can be also 'module'

attachment_preview/model/ir_attachment.py
```diff
@@ -0,0 +1,43 @@
...
+        for this in self.browse(
+                cr, uid,
+                ids if isinstance(ids, collections.Iterable) else [ids],
+                context=context):
+            extension = ''
+            if this.datas_fname:
+                filename, extension = os.path.splitext(this.datas_fname)
+                extension = extension.lstrip('.')
+            if not extension:
+                # TODO: lookup the extension via mimetype
```
pedrobaeza: I think a lot of attachments falls in this case, so it's very needed this part. https://docs.python.org/2/library/mimetypes.html can be used.

hbrunn:
Thanks Pedro for you input!

pedrobaeza:
:+1:

bwrsandman:
:+1:

PR #17, opened 2014-09-10
[ADD] document_url: A module that allows to put URLs as attached documents.

pedrobaeza:


document_url/model/document_url.py
```diff
@@ -0,0 +1,59 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2004-2010 Tiny SPRL (<http://tiny.be>).
```
gurneyalex: is this correct and up to date?

document_url/model/document_url.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        if context is None:
+            context = {}
+        if not context.get('active_model'):
+            return
+        attachment_obj = self.pool['ir.attachment']
+        for form in self.browse(cr, uid, ids, context=context):
+            url = urlparse(form.url)
+            if not url.scheme:
+                url = urlparse('%s%s' % ('http://', form.url))
```
gurneyalex, line 47: this looks weird. Why not put the 'http://' part in the first part of the string? (nit picking, I know, but...)

document_url/static/src/js/url.js
```diff
@@ -0,0 +1,46 @@
...
+                options.on_close();
+            }
+            this.dialog_stop();
+            this.inner_widget.views[this.inner_widget.active_view].controller.reload();
+            return $.when();
+        },
+    });
+
+    var _t = instance.web._t,
+    QWeb = instance.web.qweb;
```
gurneyalex: could you indent this to match `_t`? This would improve readability.

document_url/model/document_url.py
pedrobaeza, line 47: This is because you can attach not only http, but https, ftp, ftps... any valid URI. Odoo is able to redirect it to the corresponding resource.

pedrobaeza:
@gurneyalex, I have addressed your remarks, except one, that I have commented.

.travis.yml
```diff
@@ -13,10 +14,8 @@ install:
   - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly ${VERSION}
```
hbrunn: giving the version as commandline argument is deprecated from my understanding

.travis.yml
```diff
@@ -13,10 +14,8 @@ install:
...
-  - $HOME/maintainer-quality-tools/travis/travis_install_nightly ${VERSION}
 
 script:
-  - travis_run_flake8
   - travis_run_tests ${VERSION}
```
hbrunn: same here

.travis.yml
pedrobaeza: You're right. I'll update it.

document_url/model/document_url.py
```diff
@@ -0,0 +1,58 @@
...
+                attachment = {
+                    'name': form.name,
+                    'type': 'url',
+                    'url': url.geturl(),
+                    'user_id': uid,
+                    'res_id': active_id,
+                    'res_model': context['active_model'],
+                }
+                attachment_obj.create(cr, uid, attachment, context=context)
+        return {'type':  'ir.actions.act_close_wizard_and_reload_view'}
```
hbrunn, line 58: If this doesn't happen with ir.actions.act_window_close, that's a bug I think. Or am I missing something?

document_url/static/src/js/url.js
```diff
@@ -0,0 +1,46 @@
...
+        },
+    });
+
+    var _t = instance.web._t,
+        QWeb = instance.web.qweb;
+    instance.web.Sidebar.include({
+        redraw: function() {
+            var self = this;
+            this._super.apply(this, arguments);
+            self.$el.find('.oe_sidebar_add_attachment').after(QWeb.render('AddUrlDocumentItem', {widget: self}))
```
hbrunn, line 21: Why don't you inject the element by extending Sidebar in your qweb file?

document_url/model/document_url.py
pedrobaeza, line 58: I suppose you're talking that the reload view must be automatic closing wizard, doesn't it? Well, I didn't remember well if I copy/paste from google documents code, or I tried it and didn't get the view reloaded. I know that with ORM, the problem exists in https://bugs.launchpad.net/bugs/1155525, so maybe also apply to JS.

document_url/static/src/js/url.js
pedrobaeza, line 21: My knowledge in JS is very limited and I copied/pasted the code from Google documents standard module.

document_url/model/document_url.py
hbrunn, line 58: Then try to return only ir.actions.act_window_close. If we didn't reintroduce this bug, it should be fixed both in ocb and upstream by now.

pedrobaeza, line 58: I have checked it with last OCB revision and indeed the only way to get the view refreshed is with this.

pedrobaeza:
@hbrunn , can you please re-review this?

max3903:
:+1:

hbrunn:
@pedrobaeza sorry for the delay, I'm fine with the code as it is :+1:

PR #18, opened 2014-09-25
[UPD] In create method, use name field unless name_get

ehdem:
In create method, use name unless name_get to define res_name value.
In account.voucher model, the name_get return the amount, and we don't want this, we would like to get the name field.

### DIFF
--- a/document_multiple_records/document.py
+++ b/document_multiple_records/document.py
@@ -26,10 +26,38 @@ from openerp.osv import orm, fields
 class document_file(orm.Model):
     _inherit = 'ir.attachment'
 
+    def _name_get_resname(self, cr, uid, ids, object, method, context):
+        data = {}
+        context = context or {}
+        if isinstance(ids, (int, long)):
+            ids = [ids]
+        for attachment in self.browse(cr, uid, ids, context=context):
+            model_object = attachment.res_model
+            res_id = attachment.res_id
+            if model_object and res_id:
+                model_pool = self.pool.get(model_object)
+                res = model_pool.name_get(cr, uid, [res_id], context=context)
+                res_name = res and res[0][1] or False
+                # For account.voucher model, take the name field
+                if model_object == 'account.voucher':
+                    res_name = model_pool.browse(
+                        cr, uid, res_id, context=context).name
+                if res_name:
+                    field = self._columns.get('res_name', False)
+                    if field and len(res_name) > field.size:
+                        res_name = res_name[:field.size - 3] + '...'
+                data[attachment.id] = res_name
+            else:
+                data[attachment.id] = False
+        return data
+
     _columns = {
         'attachment_document_ids': fields.one2many('ir.attachment.document',
                                                    'attachment_id',
                                                    'Records'),
+        'res_name': fields.function(_name_get_resname, type='char',
+                                    size=128, string='Resource Name',
+                                    store=True),
     }
 
     def create(self, cr, uid, data, context=None):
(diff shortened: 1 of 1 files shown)


bwrsandman:
:+1:

document_multiple_records/document.py
```diff
@@ -44,7 +44,7 @@ def create(self, cr, uid, data, context=None):
                 'res_name': data.get(
                     'res_name', self.pool.get(data['res_model']).browse(
                         cr, uid, data['res_id'],
-                        context=context).name_get()[0][1]),
+                        context=context).name),
```
hbrunn: I think this must be more defensive. Every model is guaranteed to have a name_get, but 'name' is just a field that may or may not exist.
At least, you should check for existence of this field and fall back to name_get if necessary.
But in general, I'm not sure if it's a good idea to prefer a name field to name_get. There are a lot of reasons why a model would decide to have another name than 'name'. Wouldn't it be better to keep a list of known bad models ('account.voucher' for now) that get special treatment, and use name_get otherwise?

pedrobaeza: I would check with corresponding defenses (try):
- First name_get
- Then name
- After this, "model_name,id" (as reference fields do).

ehdem: @pedrobaeza : I think @hbrunn is right; It's better to add  a list of known bad models as 'account.voucher'.
We can have a model in the same time with the name field and name_get

pedrobaeza:
Why the need to have a blacklist when you can have an standard method? I repeat my list: I would check with corresponding defenses (try):
- First name_get
- Then name
- After this, "model_name,id" (as reference fields do).

ehdem:
@pedrobaeza : I agree with you, but if we take account.voucher case: 
Account.voucher has a name_get and a name field.
- The result of name_get is an amount and we don't want to save this in the res_name.
- In this case, we prefer to just save the name rather than the name_get

bwrsandman:
@pedrobaeza that's the whole point of this PR

pedrobaeza:
I would prefer to push Odoo to change account_voucher _name_get_ method than monkey-patching this...

bwrsandman:
@pedrobaeza I doubt this will ever happen, they did explicitly set the name_get to be the amount
https://github.com/odoo/odoo/commit/4f36684661a93bf791fd3d440db3960a524bd160

pedrobaeza:
OK, I see, then let's do it your way. :+1: 

Regards.

document_multiple_records/document.py
```diff
@@ -26,10 +26,38 @@
...
+                data[attachment.id] = res_name
+            else:
+                data[attachment.id] = False
+        return data
+
     _columns = {
         'attachment_document_ids': fields.one2many('ir.attachment.document',
                                                    'attachment_id',
                                                    'Records'),
+        'res_name': fields.function(_name_get_resname, type='char',
```
bwrsandman, line 58: At which point is this field used?

ehdem, line 58: This field is already defined in ir.attachment: I don't know what you really want to say ?

bwrsandman, line 58: Your commit message doesn't explain why this change was made.

When I look at the change it seems like you want to make use of the `res_name` in the `create` function, yet do do not.

It also looks like you want to factor out the `account.voucher` which is a good move, yet you left that in `create`.

Therefore, we don't know what that commit is supposed to be doing.

bwrsandman, line 58: This looks better, thanks

ehdem, line 58: Thanks you for your good comments

hbrunn:
:+1:

PR #22, opened 2014-10-24
[ADD] document_url: Module to attach URLs as documents

pedrobaeza:
Module that was accepted a few days ago on 7.0 branch (and thus not included on `__unported__` dir), migrated to new version.


hbrunn:
:+1:

PR #24, opened 2014-11-11
document_page porting to odoo 8.0

gborelli:
I started from this branch - https://github.com/savoirfairelinux/knowledge/tree/8.0-document-page - to migrate document_page to oboo 8.0

I used the new API for models and wizards
and 
I removed some parts that I didn't find in the original launchpad repository - http://bazaar.launchpad.net/~openerp/openobject-addons/7.0/files/head:/document_page/

PR #25, opened 2014-11-12
document_page_approval porting to 8.0

gborelli:
document_page_approval depends on document_page, see: https://github.com/OCA/knowledge/pull/24


hbrunn:
It looks like you're still working on this, so I tagged this PR 'work in progress'. When you're done, could you replace the tag by 'needs review'?

llacroix:
Anyone can restart travis for this one. The travis log is strange, I have no problem running this module with tests.

hbrunn:
travis restarted

bwrsandman:
Module fails to install

max3903:
@gborelli Are you planning to fix it or shall we take over ? I would like to have it merged ASAP, as we have a lot of modules depending on it. Thanks.

gborelli:
@max3903 Unfortunately I can't fix the issue related to the module installation.

dreispt:
@max3903 I still think this dependency should be made optional on the `management-system` repo...

max3903:
@dreispt you can install it and not use it (uncheck review on document page category). At least, the feature is there for those who want to use it.

dreispt:
@max3903 I agree the feature is important and should be available.
I just don't see why it needs to be automatically installed.
It bugs me that mgmt-system basic features can't be installed if the knowledge is not also available.

llacroix:
@dreispt we at least need the `document.page` but I agree that the document page approval might not be necessary.

llacroix:
I'm not sure if it's posssible to solve this issue without implementing something that looks like a hack. The problem is that odoo is using workflows of modules that aren't yet loaded.

If we run a test case once, it will work without problem. On a second run, it will try to load demo data again, the demo data will trigger a workflow in "document.page.history" in the create method. The workflow will try to execute the workflow that was previously added by by document_page_approval. Unfortunately the workflow in document_page_approval is calling a method that doesn't yet exist on the document_page_history.

I implemented a small hack on my branch yet it hope there is a better way.

bwrsandman:
Needs a rebase.

max3903:
Resumed in #30 by @llacroix

PR #30, opened 2014-11-24
[wip] 8.0 document page approval

llacroix:
Need it for some modules, trying to make it pass the build


llacroix:
I believe the error might not be related to the addons itself.

llacroix:
With the fix directly in odoo, there is no need for a hack in document_page_history

max3903:
@llacroix is it still a wip ?

llacroix:
@max3903 It's not a wip as in I solved the problem with the travis build and tests are running. One of the workflow calls a method that doesn't exist in document_page_history. During tests, it creates a document_page, that creates a document_page_history that triggers a workflow. The workflow will trigger the one in document_page_approval yet document_page_approval isn't yet loaded.

For that reasons, the tests were failing. 

Until odoo fix the way it loads workflows, we have to add the missing methods to document_page_history, or handle get_attr by ourselves to prevent KeyErrors

document_page/document_page.py
```diff
@@ -177,3 +180,14 @@ def getDiff(self, v1, v2):
...
+    def __getattr__(self, attr):
+        """Return a dummy callabale"""
+        if attr in ['_sql', 'init', '_ids']:
+            raise AttributeError
+
+        _logger.warning(
+            "Trying to access attribute %s on document_page_history",
+            attr
+        )
+        return (lambda *a, **b: None)
```
llacroix, line 193: Could we handle the workflow calling undefined method that won't be a hack...

dreispt, line 193: I don't understand the purpose of this method.

dreispt:
LGTM. Let's get this going - it's  blocking all PRs in `mgmt-system`. 
:+1:

max3903:
:+1:

document_page/document_page.py
llacroix, line 193: There is a problem in openerp, workflow will run even if modules aren't loaded during tests for example. For this reason, document_page was calling a workflow from document_page_approval which called a method that doesn't exist yet. It should be refactored differently.

PR #34, opened 2015-01-26
[ADD] attachments_to_filesystem

hbrunn:


pedrobaeza:
I think this is more a server-tool than a knowledge addon, don't you think?

StefanRijnhart:
Tested 7.0 version in good order, thanks! :+1:

attachments_to_filesystem/models/ir_attachment.py
```diff
@@ -0,0 +1,112 @@
...
+        location = self.pool['ir.config_parameter'].get_param(
+            cr, uid, 'ir_attachment.location')
+        if location:
+            # we assume the user knows what she's doing. Might be file:, but
+            # also whatever other scheme shouldn't matter. We want to bring
+            # data from the database to there
+            pass
+        else:
+            ir_model_data._update(
+                cr, uid, 'ir.config_parameter', module_name,
```
StefanRijnhart, line 46: Please set noupdate to True in this call to prevent this key to be removed upon module uninstallation, which would make all attachments unavailable.

StefanRijnhart:
Thanks! :+1: again.

pedrobaeza:
:+1:

PR #35, opened 2015-01-26
[ADD] attachments_to_filesystem

hbrunn:


pedrobaeza:
I think this is more a server-tool than a knowledge addon, don't you think?

hbrunn:
I sometimes fear server-tools degrades to some kind of modules-of-which-i-dont-know-where-to-put-them bin, so I try to only put things there that really belong. As this module deals with attachments, I think the knowledge branch is the better place. Let's see what others think

dreispt:
IMO "Knowledge" seems a good place for any attachment related modules.

pedrobaeza:
OK then

attachments_to_filesystem/models/ir_attachment.py
```diff
@@ -0,0 +1,145 @@
...
+                continue
+            ir_attachment.write(
+                cr, uid, [attachment_id],
+                {
+                    'datas': attachment_data['datas'],
+                    'db_datas': False,
+                },
+                context=context)
+            if not counter % (len(attachment_ids) / 100 or limit):
+                logging.info('moving attachments: %d done', counter)
```
gdgellatly, line 112: consider adding % or per cent here

attachments_to_filesystem/models/ir_attachment.py
```diff
@@ -0,0 +1,145 @@
...
+        attachment_ids = ir_attachment.search(
+            cr, uid, [('db_datas', '!=', False)], limit=limit, context=context)
+        logging.info('moving %d attachments to filestore', len(attachment_ids))
+        counter = 1
+        # attachments can be big, so we read every attachment on its own
+        for attachment_id in attachment_ids:
+            attachment_data = ir_attachment.read(
+                cr, uid, [attachment_id], ['datas', 'res_model'],
+                context=context)[0]
+            if attachment_data['res_model'] and not self.pool.get(
```
gdgellatly, line 98: mixed self.pool.get() syntax with self.pool[] . Maybe you need to but if you don't I always think it is a bit cleaner to just use one.

attachments_to_filesystem/models/ir_attachment.py
```diff
@@ -0,0 +1,145 @@
...
+            ir_attachment.write(
+                cr, uid, [attachment_id],
+                {
+                    'datas': attachment_data['datas'],
+                    'db_datas': False,
+                },
+                context=context)
+            if not counter % (len(attachment_ids) / 100 or limit):
+                logging.info('moving attachments: %d done', counter)
+            counter += 1
```
gdgellatly: in my own attachment migration I run it very similarly, but use enumerate in the for loop for the counter

e.g

``` python
for idx, id in enumerate(att_ids):
```

attachments_to_filesystem/models/ir_attachment.py
hbrunn, line 98: that's necessary. get() syntax for cases where the model possibly doesn't exist, [] otherwise

attachments_to_filesystem/models/ir_attachment.py
hbrunn, line 112: counter contains an amount, not a percentage

attachments_to_filesystem/models/ir_attachment.py
hbrunn: thanks, good one!

attachments_to_filesystem/models/ir_attachment.py
```diff
@@ -0,0 +1,145 @@
...
+            attachment_data = ir_attachment.read(
+                cr, uid, [attachment_id], ['datas', 'res_model'],
+                context=context)[0]
+            if attachment_data['res_model'] and not self.pool.get(
+                    attachment_data['res_model']):
+                logging.warning(
+                    'not moving attachment %d because it links to unknown '
+                    'model %s', attachment_id, attachment_data['res_model'])
+                continue
+            ir_attachment.write(
```
gdgellatly, line 104: If its going to fail it will be here.  Have you considered a try except with savepoint, so it can keep going, or at least gracefully exit if some error occurs.

attachments_to_filesystem/models/ir_attachment.py
```diff
@@ -0,0 +1,145 @@
...
+                    'datas': attachment_data['datas'],
+                    'db_datas': False,
+                },
+                context=context)
+            if not counter % (len(attachment_ids) / 100 or limit):
+                logging.info('moving attachments: %d done', counter)
+            counter += 1
+        # see if we need to create a new cronjob for the next batch
+        if ir_attachment.search(
+                cr, uid, [('db_datas', '!=', False)], limit=1,
```
gdgellatly: Holger,

You probably understand better than me why it is done this way, but why not just use a regular cron and then deactivate/delete when done?  Seems a lot of effort creating new cron jobs each time.

attachments_to_filesystem/models/ir_attachment.py
gdgellatly, line 112: Wow I read that line wrong.  My similar line looks like

``` python
if not idx % 50:
        print 'Migrated ID %d (attachment %d of %d)' % (id, idx, cnt)
```

attachments_to_filesystem/models/ir_attachment.py
hbrunn, line 104: the error that can occur is that core tries to access the model linked in res.model and it doesn't exist. That's why I check this before and continue if it's the case. All other errors should fail imho as they are unexpected and should cause a rollback

gdgellatly:
Hi Holger,

I just read through and made some very minor comments.  Unfortunately a hard one to test as I have nothing to really test it with, but compared your code to my own one off migration scripts which I know work and yours is much more thorough and safe than mine.

There a few parts I don't understand fully but it all seems to make sense.  I wouldn't worry too much about my other comments, just things I thought going along.

Cheers
Graene

attachments_to_filesystem/models/ir_attachment.py
hbrunn: hmmm, why not? I think you're right

StefanRijnhart:
Tested in good order, thanks! :+1:

attachments_to_filesystem/models/ir_attachment.py
guewen: Also beware of the special case when the search returns records which are all attached to non-existent models, it should not continue indefinitely.

hbrunn: As soon as I find time, I'll remove all the code for creating a new cronjob, set the cronjob to run daily and put a note into the readme for the user to watch the logs and deactivate the cronjob as appropriate (unfortunately, a cronjob can't deactivate itself because the scheduler has a lock on the record)

PR #36, opened 2015-01-27
[ADD] port document_choose_directory

hbrunn:
# Choose document directories

This addon enables users to choose a directory while uploading a file. In
combination with access permissions on directories, this can be used to
restrict access on certain records' attachments or simply for a better
overview.
# Configuration

Create a static directory and choose the resource model you want to use this
directory on. Assign groups to whom this directory should be accessible. When
uploading a files, users can select one of the directories they have access to,
or no directory which is the default setting.

Keep in mind that if a directory has a owner, it's only visible for this user,
that's probably not what you want.


document_choose_directory/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from . import model
+from . import controllers
```
ehdem, line 22: better to do: 
from . import (
    model,
    controllers,
)

It avoid to copy for each line "from . import"

document_choose_directory/model/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+from . import document_directory
+from . import ir_attachment
```
ehdem, line 22: same thing; Better to do and more simple to do: 
from . import (
    document_directory,
    ir_attachment,
)

hbrunn, line 22: with both I can't see how this improves anything. Can you point me to a style guide that suggests this?

document_choose_directory/model/ir_attachment.py
```diff
@@ -0,0 +1,37 @@
...
+class IrAttachment(Model):
+    _inherit = 'ir.attachment'
+
+    def read(self, cr, uid, ids, fields_to_read=None, context=None,
+             load='_classic_read'):
+        '''inject the extra field we need in the web client. This saves us a
+        couple of extra client side calls'''
+        if fields_to_read == ['name', 'url', 'type', 'create_uid',
+                              'create_date', 'write_uid', 'write_date']:
+            fields_to_read = fields_to_read + ['parent_id']
```
StefanRijnhart, line 34: Looking at the code around on_attachments_loaded, I understand why you chose a solution like this. To make this a little more robust, instead of relying on the exact contents and order, would you consider something like this

```
if fields_to_read is None:
    fields_to_read = []
if set(['name', 'url', 'type', 'create_uid', 'create_date', 'write_uid', 'write_date']) <= \
        set(fields_to_read) and 'parent_id' not in fields_to_read:
    fields_to_read.append('parent_id')
```

?

StefanRijnhart:
Thanks, a small remark inline, otherwise very nice! As for @ehdem's comment, I found https://www.python.org/dev/peps/pep-0328/ which introduces multi-line imports to solve problems making longer import statements fit Pythonic style guidelines. I don't see these problems here though so I think the current import style is fine.

document_choose_directory/model/ir_attachment.py
hbrunn, line 34: thanks, good idea. I rather don't append to the field list because we should only assume it's iterable, not that it's a list (don't ask which code passed a tuple here, but it happened). Ending up with two parent_ids isn't a problem imho

StefanRijnhart:
:+1:

document_choose_directory/__init__.py
guewen, line 22: It is totally ok to put one import per line, and it is what is done in all the existing OCA's codebase.

guewen:
:+1:

PR #40, opened 2015-02-18
6.1 attachment file type

StefanRijnhart:
# File types for attachments

Adds a content type to attachments. This is useful for instance in combination
with http://bazaar.launchpad.net/~ocb/ocb-web/6.1/revision/2524, which makes
Odoo offer downloads with the correct file type based on this field.

The document module makes this field available as well. This module does not
interfere with its functionality when both are installed.
# Installation

To install this module, you need to install python-magic in your environment
first.
# Configuration

This module comes with a daily scheduled task to provide content types for
attachments without one in small batches (instead of trying to do this for all
attachments at installation time). Depending on the number of attachments in
your system you may deactivate this task after some time. You can follow the
progress in the logs at the time that the cron job is fired.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ virtualenv:
   system_site_packages: true
 
 install:
+  - pip install python-magic
   - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
   - export PATH=${HOME}/maintainer-quality-tools/travis:${PATH}
   - travis_install_nightly ${VERSION}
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Why do you need the cron? Can't you make it with a migration script?

StefanRijnhart:
That's what the introduction says: if you have large amounts of documents, you might want to migrate gradually (and out of office hours).

pedrobaeza:
Sorry, I didn't see it. Do you think the time is too high to need this?

StefanRijnhart:
Well, it took a couple of minutes for 20000 attachments. I guess that is not too bad but if you can prevent it, why not.

pedrobaeza:
:+1:

hbrunn:
:+1: (travis error unrelated)

PR #52, opened 2015-05-05
[ADD] attachment_edit

hbrunn:
# Edit attachments

This module adds a button on the records' attachment dropdown that allows to
open the attachment's form in edit mode. This makes it convenient to edit the
attachment's properties after upload.

On the attachment's form, a field is added to move the attachment to another
record.
# Usage

Click the pencil that appears next to the delete button when you hover over
an attachment in order to edit it.

To change the resource the attachment belongs to, edit the field `Resource
reference`.


anthony-muschang:
:+1: 

But there is a display problem with the module "document_choose_directory".

display_problem

StefanRijnhart:
@hbrunn, is this compatibility with document_choose_directory something you could look at, given that you are the author of that module as well?

hbrunn:
@StefanRijnhart @anthony-muschang what do you consider the compatibility problem?

attachment_edit/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Edit attachments",
+    "version": "1.0",
```
dreispt: 8.0.1.0.0

dreispt:
I'm won't pick on the README template, and other than the version number LGTM :+1:

hbrunn:
@dreispt thanks, I updated to the current readme template anyways hoping the link to the issues will save me some of the annoying private support mails i receive about modules without it

hbrunn:
can we merge this?

attachment_edit/__openerp__.py
```diff
@@ -0,0 +1,44 @@
...
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    "name": "Edit attachments",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV",
```
pedrobaeza: Add OCA as author

attachment_edit/__openerp__.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
+##############################################################################
```
pedrobaeza: Can you use short headers?

attachment_edit/static/src/css/attachment_edit.css
```diff
@@ -0,0 +1,22 @@
+.openerp .oe_sidebar .oe_dropdown_menu li .oe-sidebar-attachment-edit
+{
+    position: absolute;
```
pedrobaeza, line 3: There's no other option that to put an absolute position? Can you insert the button in the flow via a JQuery command?

hbrunn, line 3: unfortunately not, because the whole setup from core only allows absolute placements.
I needed the same thing for https://github.com/OCA/knowledge/blob/9.0/attachment_preview/static/src/css/attachment_preview.css, which is why those two modules can coexist in peace. But that doesn't scale at all or course.

What we need to do this right is in my opinion one module that moves core's delete button (the cross) into a qweb template, then all subsequent modules can simply add their buttons to this template. I'm quite sure to remember you, me and some other person had a similar discussion on some other PR, but I can't find it back now.

How about this: I hereby pledge that if you let the absolute placement slip here, I'll provide the base module described above and appropriate PRs for `attachment_edit` and `attachment_preview` as soon as a PR for a third button-to-attachment-adding module is PRed

pedrobaeza, line 3: Yes, the discussion was in https://github.com/OCA/knowledge/issues/87. OK, let's keep this and with the next added button, we make the move.

pedrobaeza:
:+1:

dreispt:
Two OKs, looks ready to merge.

PR #55, opened 2015-07-13
[ADD] document_no_unique_filenames

hbrunn:
# No unique filenames in directories

This module drops the uniqueness constraint on filenames within the same directory. This constraint only makes sense if the directories are exported to actual filesystems, which is not the case with the two standard storage options within odoo.


pedrobaeza:
Isn't better to do this on OCB?

hbrunn:
Well, it might be interesting for non-ocb-users too. Another thing is that if there's some storage implementation out there that needs unique filenames, I'd rather not break it (even though I don't know of such a thing)

document_no_unique_filenames/models/ir_attachment.py
```diff
@@ -0,0 +1,37 @@
...
+    _inherit = 'ir.attachment'
+
+    def _register_hook(self, cr):
+        cr.execute(
+            "select c.id from ir_model_constraint c, ir_model m "
+            "where c.model=m.id and m.model='ir.attachment' and "
+            "c.name='ir_attachment_filename_unique'")
+        # this function does all the magic, but only if we select all
+        # references to the constraint above, meaning all inheriting
+        # module that also declare ir.attachment
```
pedrobaeza, line 34: s/module/modules

pedrobaeza:
Minor comment, but :+1:

max3903:
:+1:

PR #56, opened 2015-07-21
[UPD] use viewerjs 0.5.7

hbrunn:
This is just a library update, the only intersting change is in `attachment_preview/static/src/js/viewerjs_tweaks.js`


pedrobaeza:
Tried on runbot and working as expected, so as this is a technical update, I fast-track it.

PR #61, opened 2015-08-19
[FIX] - document-page-approval: Fixes #60 url not valid when website module is installed.

### LINKED ISSUE #60: Module document_page_approval: URL does not match when website is installed.
The url does not work if website module is installed
http://localhost:8001/#db=test&id=1&view_type=form&model=document.page.history

It needs to be
http://localhost:8001/web#db=test&id=1&view_type=form&model=document.page.history
with /web

ghost:


pedrobaeza:
I think there's a method to get the URL prefix to avoid this manual adjustment. Can you please check it?

ghost:
> I think there's a method to get the URL prefix to avoid this manual adjustment. Can you please check it?

Hi @pedrobaeza 

As far as I know Odoo does it like this in auth_signup module res_users.py

``` python
base_url = self.pool.get('ir.config_parameter').get_param(cr, uid, 'web.base.url')
...
res[partner.id] = urljoin(base_url, "/web/%s?%s" % (route, werkzeug.url_encode(query)))
```

Do you know where can I find that method to avoid including it manually?

pedrobaeza:
OK, I thought it exists. Then :+1:

max3903:
:+1:

PR #68, opened 2015-09-16
[ADD] allow to set maximum number of records to reindex

hbrunn:
This is for situations where the amount of attachments is so huge we can't do it in one transaction


dreispt:
:+1:

pedrobaeza:
:+1:

Issue #54, opened 2015-05-06
[8.0] TypeError: function lambda at 0x7f1f2cb562a8 is not JSON serializable


alanljj:
after installed, when click menu Marketing-MassMailing, the message pop

Odoo Server Error

Traceback (most recent call last):
  File "/opt/odoo/openerp/http.py", line 530, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/opt/odoo/openerp/http.py", line 581, in dispatch
    return self._json_response(result)
  File "/opt/odoo/openerp/http.py", line 519, in _json_response
    body = simplejson.dumps(response)
  File "/home/odoo/odoo/lib/python2.7/site-packages/simplejson/__init__.py", line 370, in dumps
    return _default_encoder.encode(obj)
  File "/home/odoo/odoo/lib/python2.7/site-packages/simplejson/encoder.py", line 269, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/home/odoo/odoo/lib/python2.7/site-packages/simplejson/encoder.py", line 348, in iterencode
    return _iterencode(o, 0)
  File "/home/odoo/odoo/lib/python2.7/site-packages/simplejson/encoder.py", line 246, in default
    raise TypeError(repr(o) + " is not JSON serializable")
TypeError: <function <lambda> at 0x7f1f2cb562a8> is not JSON serializable

pedrobaeza:
What module have you installed?

alanljj:
the module of document_page

akram1:
Same problem more details as follows:

Error on document_page

Odoo 8 (build 26/7/2015) Ubuntu 14.04, Postgres 9.4.1, werkzeug 0.9.6, Python 2.7.8

On a clean Odoo DB, installing document_page then installing crm_mass_mailing cause marketing manu top menu to stop working and make this error :
  File "/media/akram/Data/projects/servers/Odoo/odoo8/openerp/http.py", line 537, in _handle_exception
    return super(JsonRequest, self)._handle_exception(exception)
  File "/media/akram/Data/projects/servers/Odoo/odoo8/openerp/http.py", line 588, in dispatch
    return self._json_response(result)
  File "/media/akram/Data/projects/servers/Odoo/odoo8/openerp/http.py", line 526, in _json_response
    body = simplejson.dumps(response)
[...]

PR #82, opened 2015-12-30
Fix TypeError: function lambda at 0x7f1f2cb562a8 is not JSON serializable

max3903:
#54


max3903:
ping @alanljj, @akram1, @whoople, @ccarly

dreispt:
LGTM :+1:

PR #97, opened 2016-08-03
[FIX][document_sftp] Protect import

yajo:
This module has an unprotected import that breaks the system when `paramiko` is not installed.

@Tecnativa


hbrunn:
:+1:

lasley:
👍

PR #100, opened 2016-08-25
[IMP] attachment_preview: improve performance for large files with filenames

tremlin:
This PR rewrites `get_binary_extension` so that it uses a first pass to determine the extension from the filename only whenever possible. This avoids to load big files into memory if only the filename is needed.

### DIFF
--- a/attachment_preview/model/ir_attachment.py
+++ b/attachment_preview/model/ir_attachment.py
@@ -61,8 +61,13 @@ class IrAttachment(Model):
                     hasattr(magic, 'MAGIC_MIME_TYPE') and
                     magic.MAGIC_MIME_TYPE or magic.MAGIC_MIME)
                 ms.load()
-                mimetype = ms.buffer(
-                    base64.b64decode(this[binary_field]))
+                if model == self._name and binary_field == 'datas'\
+                        and this.store_fname:
+                    mimetype = ms.file(
+                        this._full_path(cr, uid, this.store_fname))
+                else:
+                    mimetype = ms.buffer(
+                        base64.b64decode(this[binary_field]))
             except ImportError:
                 (mimetype, encoding) = mimetypes.guess_type(
                     'data:;base64,' + this[binary_field], strict=False)
--- a/attachment_preview/model/ir_attachment.py
+++ b/attachment_preview/model/ir_attachment.py
@@ -32,35 +32,45 @@ class IrAttachment(Model):
             self, cr, uid, model, ids, binary_field, filename_field=None,
             context=None):
         result = {}
-        for this in self.pool[model].browse(
-                cr, uid,
-                ids if isinstance(ids, collections.Iterable) else [ids],
-                context=context):
-            if not this.id:
-                result[this.id] = False
-                continue
-            extension = ''
-            if filename_field and this[filename_field]:
-                filename, extension = os.path.splitext(this[filename_field])
-            if not this[binary_field]:
-                result[this.id] = False
-                continue
-            if not extension:
-                try:
-                    import magic
-                    ms = magic.open(
-                        hasattr(magic, 'MAGIC_MIME_TYPE') and
-                        magic.MAGIC_MIME_TYPE or magic.MAGIC_MIME)
-                    ms.load()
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Are you sure magic reads whole file? AFAIK, it only reads first bytes.

tremlin:
Magic usually reads only a few bytes, however not always at the beginning (http://linux.die.net/man/5/magic). The core problem is not really with magic but with Odoo, which provides AFAIK no fast random file access.

pedrobaeza:
OK, I let @hbrunn to give his opinion on this.

attachment_preview/model/ir_attachment.py
```diff
@@ -32,35 +32,47 @@ def get_binary_extension(
...
+                if this[filename_field]:
+                    filename, extension = os.path.splitext(
+                        this[filename_field])
+                if this[binary_field]:
+                    result[this.id] = extension
+                else:
+                    result[this.id] = False
+        # Seconds pass for all attachments which have to be loaded fully
+        #  to get the extension from the content
+        ids_to_browse = [_id for _id in ids_to_browse if _id not in result]
```
hbrunn, line 61: how can any id be left here in the happy flow? That is, don't you effectively disable mime type tests?

hbrunn:
as for random access: Since version 8, odoo stores attachments on the filesystem by default. So I think your patch would be a lot better if you check for this the way core does it in https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/ir/ir_attachment.py#L161 and just read from the file system if you can via https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/ir/ir_attachment.py#L100. This should solve your problem while still using the magic and mimetype logic

tremlin:
Thanks for the review. The logic change was not intended this way and I've corrected it. That is, if no extension could be obtained from the filename, magic will be tried on the file.

Regarding passing the storage filename to magic directly: It works only for attachments in Odoo 8 and not for binary fields in general (only Odoo 9 has the attachment attribute of `binary`). I implemented the change in a second commit, but I am not sure how helpful it is in general.
The application that I had in mind is exactly the opposite: arbitrary binary fields which are stored in the database :)

attachment_preview/model/ir_attachment.py
```diff
@@ -32,35 +32,50 @@ def get_binary_extension(
...
-                    ms = magic.open(
-                        hasattr(magic, 'MAGIC_MIME_TYPE') and
-                        magic.MAGIC_MIME_TYPE or magic.MAGIC_MIME)
-                    ms.load()
+        # First pass: load fields in bin_size mode to avoid loading big files
+        #  unnecessarily.
+        if filename_field:
+            context_bin_size = dict(context or {})
+            context_bin_size['bin_size'] = True
+            ids_to_browse = ids if isinstance(ids, collections.Iterable) \
```
hbrunn: move this out of the conditional to avoid possible crashes in line 55 below

tremlin: Thanks. Don't change code after testing...

attachment_preview/model/ir_attachment.py
```diff
@@ -32,35 +32,50 @@ def get_binary_extension(
...
-
-            result[this.id] = (extension or '').lstrip('.').lower()
+            except ImportError:
+                (mimetype, encoding) = mimetypes.guess_type(
+                    'data:;base64,' + this[binary_field], strict=False)
+            extension = mimetypes.guess_extension(
+                mimetype.split(';')[0], strict=False)
+            result[this.id] = extension
+        for _id in result:
+            result[_id] = (extension or '').lstrip('.').lower()
```
hbrunn: was this here before? This won't crash most of the time, but assign the last extension we got to all requested records.

Maybe add a test for this function so that we're sure it behaves correctly?

tremlin:
Sorry for the hot action yesterday.
I tested the now corrected code with both cases (extension from filename and file magic).

hbrunn (approved):

pedrobaeza (approved):
Following Holger's expert approval.

PR #101, opened 2016-09-29
[FIX][document_url] Consistent styling.

yajo:
Without this patch, when you hover on the menu item, its background color does not change, and its cursor is a text one:
consejo emergente_002

With this patch, the menu item styling matches those available in all surrounding menus:
consejo emergente_003

@Tecnativa

### DIFF
--- a/document_url/static/src/css/url.css
+++ b/document_url/static/src/css/url.css
@@ -1,3 +1,0 @@
-.oe_url_attachment{
-    padding: 3px 20px;
-}
--- a/document_url/static/src/xml/url.xml
+++ b/document_url/static/src/xml/url.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <templates id="template" xml:space="preserve">
 <t t-name="AddUrlDocumentItem">
-     <li class="oe_sidebar_add_url"><span class="oe_url_attachment">Add URL...</span></li>
+    <li class="oe_sidebar_add_url">
+        <a class="oe_file_attachment">Add URL...</a>
+    </li>
 </t>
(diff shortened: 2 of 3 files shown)


hbrunn (approved):
the only argument I see for the extra class is to be able to grab the element via a selector, but that's already possible with the `li`'s class

pedrobaeza (approved):

lasley (approved):
code+functional 👍

JonathanNEMRY:
Hello, @lasley did you click on add url? When I go to runbot2.odoo-community.org and take the last instance including the modifications, I have this;
selection_008
selection_009

And the previous built instance seems to work fine...any ideas?

yajo:
Indeed it's introduced by the change from span to a. Gotta fix that, thanks

sbidoul:
@pedrobaeza @Yajo can we revert this until this is resolved properly?

pedrobaeza:
No, we should work in https://github.com/OCA/knowledge/pull/102 for fixing ASAP instead, as also the error is not blocking (is annoying, but not blocking).

yajo:
Yes, #102 fixes this, no need to do any additional work.

sbidoul:
@Yajo @pedrobaeza I disagree. Before this, the module was working fine in 9c and 9e with a minor look and feel issue. Now it creates a stack trace, that's a bug. As #102 makes the module incompatible with 9e it will take time to resolve properly.

pedrobaeza:
I think this is not too much time, but we don't have access right now to 9e stack to try, so please provide us with the patch. Meanwhile, you can freeze in the previous commit your deployment.

PR #102, opened 2016-09-30
[FIX][document_url] Remove error when clicking on add URL

yajo:
BTW, completely refactored :wink:. It turns out Odoo's framework is not that bad when you discover it. It's just not documented.

Fixes https://github.com/OCA/knowledge/pull/101#issuecomment-250673496

@Tecnativa @JonathanNEMRY @pedrobaeza @hbrunn @lasley

### DIFF
--- a/document_url/static/src/xml/url.xml
+++ b/document_url/static/src/xml/url.xml
@@ -2,7 +2,8 @@
 
 <templates>
     <t t-extend="Sidebar">
-        <t t-jquery=".oe_sidebar_add_attachment" t-operation="after">
+        <!-- oe_ for Odoo CE, and o_ for Odoo EE -->
+        <t t-jquery=".oe_sidebar_add_attachment, .o_sidebar_add_attachment" t-operation="after">
             <li class="oe_sidebar_add_url">
                 <a class="oe_file_attachment"
                     data-section="url_doc"
--- a/document_url/static/src/xml/url.xml
+++ b/document_url/static/src/xml/url.xml
@@ -1,8 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<templates id="template" xml:space="preserve">
-<t t-name="AddUrlDocumentItem">
-    <li class="oe_sidebar_add_url">
-        <a class="oe_file_attachment">Add URL...</a>
-    </li>
-</t>
+
+<templates>
+    <t t-extend="Sidebar">
+        <t t-jquery=".oe_sidebar_add_attachment" t-operation="after">
+            <li class="oe_sidebar_add_url">
+                <a class="oe_file_attachment"
+                    data-section="url_doc"
+                    data-index="0">Add URL...</a>
+            </li>
+        </t>
+    </t>
 </templates>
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):
Tried on runbot

lasley (approved):
Ohhhh this is nice 👍

JonathanNEMRY:
Hello @Yajo 
Thanks for the fix. The code seems more cleaner than before.

However, I can't approve this PR as the previous version was supported into odoo enterprise and this one is no more.

The previsous js selector was applied on two different classes : oe_ and o_ and then it was possible to select a comon anchor whatever you was into community or enterprise.

This workaround was made because of the difference on the xml data structure level between both odoo version

So after tested it on 9e it appears that the introduced code does not seem to keep this compatibility and and consequently include an implicit regression.

Can you please provide a solution for this problem?

Jonathan

yajo:
AFAIK `oe_` prefix stands for _OpenERP_ and it was the commonly used in v7 times. Since v8, they started using `o_` in some places, coming from _Odoo_.

It's just a supposition, since there has never been an official statement on this AFAICR.

In any case, [web client incompatibilities are expected](https://www.odoo.com/es_ES/groups/community-59/community-20344972?mode=thread&date_begin=&date_end=), given the enterprise edition has a different one. I have no enterprise instance to test it, so I guess we have 3 options:
1. Merge this, and after that, you can send a new PR adding Enterprise compatibility.
2. Block this and wait until you send me a PR adding it.
3. Develop a _glue_ addon between this one and `web_enterprise`.

Personally I don't think option 2 is fair, but I'm not OCA's owner...

sbidoul:
@Yajo I read elsewhere that incompatibilities between 9c and 9e web modules were considered as bugs by Odoo. So far it as always been possible to make compatible modules. 

AFAIK, the previous version of document_url (before #101) was working fine on both version albeit with a minor style issue.

So I think it's fair to try to preserve the compatibility. I mean, if you had been the first to migrate document_url, it would be normal to leave to other the burden to ensure compatibility with 9e. But in this case, it was migrated by @JonathanNEMRY who made the effort to have it compatible so I think you should at least take into account his review comments.

That said, I understand very well it is a complex issue as not everyone has access to 9e.

document_url/static/src/xml/url.xml
```diff
@@ -1,8 +1,13 @@
...
-<templates id="template" xml:space="preserve">
-<t t-name="AddUrlDocumentItem">
-    <li class="oe_sidebar_add_url">
-        <a class="oe_file_attachment">Add URL...</a>
-    </li>
-</t>
+
+<templates>
+    <t t-extend="Sidebar">
+        <t t-jquery=".oe_sidebar_add_attachment" t-operation="after">
```
pedrobaeza: @Yajo, can you select here by the 2 selectors mentioned by @JonathanNEMRY ?

yajo:
Could somebody try on Enterprise with last patch?

pedrobaeza:
Merging as this probably works. If not, please make the proper PR.

JonathanNEMRY:
Hi @pedrobaeza 
sorry for the delay! (I had to retry on oca/community last day but it was pending all the entire day and I forggot to do it the next)
Long story made short we made a step forward with the enterprise version (the link appears without any errors) but unfortunatly we also made a step back with both version. Actually the link appears into all menu tabs see
selection_023
selection_022

Mayebe a last effort before it's ok.

pedrobaeza:
@Yajo, can you please check?

PR #104, opened 2016-10-04
[ADD] OCA codesprint fixes

hbrunn:
- a few bugfixes, including the grave one that everyone read with admin
  permissions
- write/create attachment support

### DIFF
--- a/document_sftp/document_sftp_server.py
+++ b/document_sftp/document_sftp_server.py
@@ -19,6 +19,7 @@ class DocumentSFTPServer(ServerInterface):
             if not user:
                 return AUTH_FAILED
             user.sudo(user.id).check_credentials(password)
+            self.env = self.env(user=user.id)
             return AUTH_SUCCESSFUL
         except AccessDenied:
             pass
@@ -37,6 +38,7 @@ class DocumentSFTPServer(ServerInterface):
                     'Ignoring key of unknown type for line %s', line)
                 continue
             if RSAKey(data=decodebytes(key_data)) == key:
+                self.env = self.env(user=user.id)
                 return AUTH_SUCCESSFUL
         return AUTH_FAILED
 
--- a/document_sftp/models/document_sftp_root.py
+++ b/document_sftp/models/document_sftp_root.py
@@ -63,6 +63,11 @@ class DocumentSFTPRoot(models.AbstractModel):
         raise NotImplementedError()
 
     @api.model
+    def _remove(self, path):
+        """Return file attributes"""
+        raise NotImplementedError()
+
+    @api.model
     def _lstat(self, path):
         """Return attributes about a link"""
         return self._stat(path)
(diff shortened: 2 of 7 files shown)


document_sftp/document_sftp_server.py
```diff
@@ -19,6 +19,7 @@ def check_auth_password(self, username, password):
             if not user:
                 return AUTH_FAILED
             user.sudo(user.id).check_credentials(password)
+            self.env = self.env(user=user.id)
```
pedrobaeza, line 22: Why not `self.sudo(user)`?

document_sftp/document_sftp_server.py
```diff
@@ -37,6 +38,7 @@ def check_auth_publickey(self, username, key):
                     'Ignoring key of unknown type for line %s', line)
                 continue
             if RSAKey(data=decodebytes(key_data)) == key:
+                self.env = self.env(user=user.id)
```
pedrobaeza, line 41: The same as above

document_sftp/models/document_sftp_root.py
```diff
@@ -59,6 +63,18 @@ def _open(self, path, flags, attr):
         raise NotImplementedError()
 
     @api.model
+    def _remove(self, path):
+        """Return file attributes"""
+        raise NotImplementedError()
```
pedrobaeza, line 68: Is it recommended to return an error at this level? Shouldn't we return a forbidden code or similar?

document_sftp/models/document_sftp_root.py
```diff
@@ -59,6 +63,18 @@ def _open(self, path, flags, attr):
...
+    @api.model
     def _lstat(self, path):
         """Return attributes about a link"""
         return self._stat(path)
+
+    @api.model
+    def _split_path(self, path):
+        """Return a list of normalized and stripped path components"""
+        # TODO: normalization
+        path = path.strip('/')
```
pedrobaeza, line 79: Shouldn't you use path separator from `os.path`?

document_sftp/document_sftp_server.py
hbrunn: note that the files which are located in the module's root (so not in `models`) are not models, but implementations of the respective interfaces that need to be implemented for paramiko

document_sftp/models/document_sftp_root.py
hbrunn: no, the ssh protocol defines the path separator as bein a slash: https://tools.ietf.org/id/draft-ietf-secsh-filexfer-13.txt (section 6, file names)

document_sftp/models/document_sftp_root.py
hbrunn: good question. I don't really have an opinion about this, the good thing about the exception is that it's quite clear what's going on, whereas a standard return value might also point to a permission issue or the like

document_sftp/document_sftp_server.py
pedrobaeza: OK, thanks for the explanation

document_sftp/models/document_sftp_root.py
pedrobaeza: OK, we keep as is right now.

pedrobaeza:
Can you please review Travis status?

hbrunn:
this module never went to production, and should be redesigned (ie own workers instead of threads) if ever this was picked up again

PR #110, opened 2016-11-10
8.0 document rtf index

NL66278:
Proper indexing of rtf documents. Can prevent huge performance problems by not indexing embedded images.


document_rtf_index/README.rst
```diff
@@ -0,0 +1,37 @@
+Index rtf documents
```
hbrunn, line 5: update this to the current template of the oca readme. Main difference is that all links should be https now

document_rtf_index/std_index.py
```diff
@@ -0,0 +1,61 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+import StringIO
+
+from pyth import document
```
hbrunn: protect all non-standard imports

document_rtf_index/std_index.py
```diff
@@ -0,0 +1,61 @@
...
+    def _doIndexContent(self, content):
+        """Just get text contents of rtf file."""
+        s = StringIO.StringIO(content)
+        r = Rtf15Reader.read(s)  # r will be pyth.document.Document
+        s.close()
+        w = PlaintextWriter.write(r)  # w will be cStringIO.StringO
+        result = _to_unicode(w.getvalue())
+        return result
+
+cntIndex.register(RtfDoc())
```
hbrunn: this means that your code will be used for all databases, not just for the one(s) where the module is installed. Better do this in `_register_hook` of some abstract model, or `ir.attachment` itself

document_rtf_index/tests/test_rtf_index.py
```diff
@@ -0,0 +1,28 @@
...
+            'test_files',
+            'test_with_cat_image.rtf'
+        )
+        rtf_file = open(rtf_path, 'rb').read().encode('base64')
+        att1 = self.env['ir.attachment'].create({
+            'name': 'test_with_cat_image.rtf',
+            'datas_fname': 'test_with_cat_image.rtf',
+            'datas': rtf_file,
+        })
+        self.assertEqual(att1.file_type, 'application/rtf')
```
hbrunn, line 29: this is platform dependent. Some distributions return `text/rtf` here, which doesn't seem wrong: https://en.wikipedia.org/wiki/Rich_Text_Format

hbrunn (changes requested):

document_rtf_index/tests/test_rtf_index.py
NL66278: Not wrong, neither is application/rtf. att1.file_type will be set to the first mime type returned by _getMimeTypes. So in this case not dependent on platform. The test is usefull to make sure it has really been indexed by the rtf indexer, and not by the default one.

hbrunn: thanks for the explanation

hbrunn:
an, and in order to get the tests right, pip install your package in `.travis.yml`, and for runbot, we have to ask @gurneyalex nicely to add the package pyth: https://pypi.python.org/pypi/pyth

document_rtf_index/std_index.py
NL66278: Trouble doing this. Seems _register_hook is only called after tests have been run. Leading to test failure now...

NL66278: Looking at pre_init hooks now...

NL66278: No: only runs on new install

NL66278:
I think I added al requested changes now.

document_rtf_index/std_index.py
NL66278: In the end just call _register_hook manually from test. No problem as double registration when tests are run will just replace already registered Rtf indexer class by itself.

gurneyalex:
@hbrunn done.

gurneyalex (approved):

hbrunn (approved):

hbrunn:
thanks @gurneyalex !

PR #116, opened 2016-12-21
[WIP] [BKP] Backport 8.0 rtf document indexer.

NL66278:
For the moment still need to test.


NL66278:
Not clear to me why runbot gives error. Hardly any log...

hbrunn:
there was a time when runbot errored out randomly when we stepped over to dockerized runbot. I restarted it to see if that was the problem

PR #118, opened 2016-12-23
Porting of knowledge to v10

LeartS:


knowledge/__manifest__.py
```diff
@@ -1,38 +1,21 @@
...
-##############################################################################
+# Copyright 2016 MONK Software
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Knowledge Management System",
-    "version": "9.0.1.1.0",
-    "depends": ["base_setup"],
-    "author": "OpenERP SA, Odoo Community Association (OCA)",
-    "category": "Hidden/Dependency",
+    "version": "10.0.1.1.0",
```
pedrobaeza: 10.0.1.0.0

knowledge/__manifest__.py
```diff
@@ -1,38 +1,21 @@
...
+    "author": "MONK Software, Odoo Community Association (OCA)",
```
max3903: @LeartS Please keep `OpenERP SA` in the author.

knowledge/__manifest__.py
```diff
@@ -1,38 +1,21 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2016 MONK Software
```
max3903: @LeartS Copyrights belong to Tiny SPRL

knowledge/models/res_config.py
```diff
@@ -34,28 +15,3 @@ class KnowledgeConfigSettings(models.TransientModel):
...
-        help='Connect Odoo with a CMIS compatible server to attach files\n'
-        'to an Odoo record.\n'
-        '- This installs the module cmis_read.'
-    )
-
-    module_cmis_write = fields.Boolean(
-        'Store attachments in an external DMS instead of the Odoo Filestore',
-        help='Connect Odoo with a CMIS compatible server to store files.\n'
-        '- This installs the module cmis_write.'
-    )
```
max3903, line 61: @LeartS Please keep those fields. Even if the modules are not migrated, it gives them visibility.

max3903 (changes requested):

knowledge/models/res_config.py
dreispt: Maybe make them readonly until they are ported?

knowledge/security/ir.model.access.csv
```diff
@@ -1,3 +0,0 @@
-id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
-access_ir_attachment_sale_manager,ir_attachment manager,base.model_ir_attachment,base.group_sale_manager,1,1,1,1
-access_ir_attachment_partner_manager,ir_attachment_partner manager,base.model_ir_attachment,base.group_partner_manager,1,1,1,1
```
dreispt, line 3: Why were ACLs removed?

dreispt:
@LeartS will you address my comments?

knowledge/security/ir.model.access.csv
LeartS: Because:

* the groups didn't make any sense; why should attachments be restricted to sale managers and partner managers?
* the `base.group_sale_manager` group doesn't exist anymore, it has been moved to the `sales_team` module. I don't see why the `knowledge` module should depend on `sales_team`
* attachments already implement a custom logic for permissions: attachments linked to a record (so, all "user" attachments, which are the main focus of this module) can be accessed by the same people who can access the linked record. That logic is much better and in my opinion, sufficient.

max3903 (approved):

dreispt (approved):

PR #120, opened 2017-01-05
[MIG] document_page: Migration to version 10.0

max3903:
Depends on #118


document_page/wizard/document_page_create_menu.xml
```diff
@@ -1,41 +1,40 @@
...
-                </form>
-            </field>
-        </record>
 
+    <!-- Create Menu From view -->
+    <record id="view_wiki_create_menu" model="ir.ui.view">
+        <field name="name">Create Menu</field>
+        <field name="model">document.page.create.menu</field>
+        <field name="arch" type="xml">
+            <form string="Create Menu" version="7.0">
```
hbrunn: the `string` and `version` attributes can go away I think

document_page/wizard/document_page_show_diff.xml
```diff
@@ -1,39 +1,38 @@
...
-                    </footer>
-                </form>
-            </field>
-        </record>
+    <!-- Create Index Form view  -->
+    <record id="view_wiki_show_diff" model="ir.ui.view">
+        <field name="name">Show Difference</field>
+        <field name="model">wizard.document.page.history.show_diff</field>
+        <field name="arch" type="xml">
+            <form string="Difference" version="7.0">
```
hbrunn: remove `string` and `version` attribute

hbrunn (changes requested):
only small nits

document_page/__manifest__.py
```diff
@@ -43,7 +32,5 @@
     'demo': [
         'demo/document_page.xml'
     ],
-    'installable': False,
```
dreispt, line 46: It's recommended to keep the `'installable': True` flag here.

document_page/__manifest__.py
```diff
@@ -43,7 +32,5 @@
...
-    'auto_install': False,
     'css': ['static/src/css/document_page.css'],
```
dreispt: Is this a valid manifest key?

document_page/models/document_page.py
```diff
@@ -76,26 +60,26 @@ class DocumentPage(models.Model):
...
-        select=True,
+        index=True,
         readonly=True)
 
     write_uid = fields.Many2one(
         'res.users',
         "Last Contributor",
-        select=True,
+        index=True,
         readonly=True
```
dreispt, line 78: Open question: do we really need to index any or all these these fields?

document_page/views/document_page.xml
```diff
@@ -1,309 +1,309 @@
...
-            <field name="priority">100</field>
-            <field name="arch" type="xml">
-                <tree string="Document Page">
-                    <field name="name"/>
-                    <field name="write_uid"/>
-                    <field name="write_date"/>
-                </tree>
-            </field>
-        </record>
+    <menuitem name="Pages"
```
dreispt, line 4: This is the top menu ("App") name.
Should we rather name it "Documents" or "Knowledge" ?

document_page/__manifest__.py
hbrunn: who recommends that? I usually ask people to remove keys when set to the default value

document_page/__manifest__.py
hbrunn: no, not since 7.0

document_page/__manifest__.py
pedrobaeza: This is a special case due to the OCA migrate_branch script that was not covered in the regexp expressions, so it could lead to broken manifest files, but I think @StefanRijnhart fixed it when he adapted the script for v10, so it's no more an issue (but needs confirmation by his part).

StefanRijnhart: Yes, the script now detects if the key is not present and adds it in that case. So removing it is harmless.

dreispt: Personally, keeping the install key makes it much easier to use grep to list installable/non installable modules. 
But that is just a personal preference and I won't pick on this then.

document_page/__manifest__.py
dreispt: Yes, please replace this by an XML extending the web client assets to add the intended CSS file.

document_page/views/document_page.xml
dreispt: @max3903 what do you think?

max3903: @dreispt I am fine with `Knowledge`.

max3903:
Thanks for your review everyone. I will fix the module today or tomorrow.

document_page/views/document_page.xml
max3903: @dreispt I double-checked and this is not the top menu name. It's inheriting from `knowledge.menu_document`

max3903:
I believe everything is fixed here.

document_page/README.rst
```diff
@@ -40,7 +40,7 @@ Bug Tracker
 
 Bugs are tracked on `GitHub Issues <https://github.com/OCA/Knowledge/issues>`_.
 In case of trouble, please check there if your issue has already been reported.
-If you spotted it first, help us smashing it by providing a detailed and welcomed feedback `here <https://github.com/OCA/knowledge/issues/new?body=module:%20document_page%0Aversion:%209.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback `here <https://github.com/OCA/knowledge/issues/new?body=module:%20document_page%0Aversion:%2010.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
lasley: Please update to new readme template

lasley (approved):
One comment inline, but nothing major. Code + functional 👍 

Looks like there's a warning in Runbot though:

```
2017-01-19 20:42:19,767 211 WARNING openerp_test odoo.addons.base.ir.ir_model: Creating the ir.model.data group_document_user in module base instead of knowledge.
```

Curious - is there a reason this security is being created in a module that we don't own?

max3903:
@lasley Good catch! Thanks. All fixed.

StefanRijnhart (approved):
Tested on runbot. Thanks!

max3903:
@hbrunn Can you update your review here? Thank you!

dreispt (approved):

dreispt:
Holger's concerns were fixed. Merging.

PR #122, opened 2017-01-19
[MIG] document_page_approval

max3903:
- [x] Depends #120


dreispt (approved):
It's a clean port.

max3903:
@eugen-don or @LeartS any of you can review this pull request?

Thanks!

document_page_approval/data/email_template.xml
```diff
@@ -1,20 +1,17 @@
...
-            <field name="body_html">
-                <![CDATA[
+    <record id="email_template_new_draft_need_approval" model="mail.template">
+        <field name="name">Automated new draft need approval Notification Mail</field>
+        <field name="email_from">${object.create_uid.company_id.email or 'noreply@localhost.com'}</field>
+        <field name="subject">New version of "${object.page_id.name}" to approve</field>
+        <field name="email_to">${object.get_approvers_email}</field>
+        <field name="model_id" ref="model_document_page_history"/>
+        <field name="auto_delete" eval="True"/>
+        <field name="lang">${object.create_uid.partner_id.lang}</field>
```
LeartS, line 12: Shouldn't it be in the approver language, not the page author's one?

LeartS (approved):
Quick code review, looks good 👍

document_page_approval/data/email_template.xml
max3903: @LeartS There could be many approvers with different languages. If they need to approve the content, I assumed they can at least read in the language of the creator.

LeartS: Yes, the creator language is definitely ok too, I was just wondering if there was a way to send each mail in the recipient language (isn't there a way to get the recipient in the evaluation context / variables?).

Also note that the author language is not necessarily the document language, imagine for example a global company where each employee sets their own language in their preferences, but they all contribute to documentation in english.

Anyway, I'm nitpicking. The author language is ok.

eugen-don:
seems youre all done with this. sorry for the late notice...

PR #124, opened 2017-02-03
[NEW] New module: attachment_attach_non_editable

LeartS:
In odoo, by default, you can't add attachments to a record if the form view
is set as not editable (``edit='false'`` in the form tag)

This module adds a new form view attribute, ``attach=``, which lets you
enable attachment uploads even in non-editable forms.

### DIFF
--- a/attachment_attach_non_editable/README.rst
+++ b/attachment_attach_non_editable/README.rst
@@ -1,0 +1,56 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+===============================
+Force enable attachments upload
+===============================
+
+In odoo, by default, you can't add attachments to a record if the form view
+is set as not editable (``edit='false'`` in the form tag)
+
+This module adds a new form view attribute, ``attach=``, which lets you
+enable attachment uploads even in non-editable forms.
+
+Usage
+=====
+
+Add the ``attach=true`` attribute to the form view tag which of uneditable views
+(normal editable views already allow attaching a document and there is no need
+for this attribute).
+
+
+Bug Tracker
+===========
+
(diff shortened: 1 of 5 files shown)


attachment_attach_non_editable/README.rst
```diff
@@ -0,0 +1,56 @@
...
+enable attachment uploads even in non-editable forms.
+
+Usage
+=====
+
+Add the ``attach=true`` attribute to the form view tag which of uneditable views
+(normal editable views already allow attaching a document and there is no need
+for this attribute).
+
+
```
max3903: Can you add the link to Runbot?

max3903 (changes requested):
Small change. 

:+1:

attachment_attach_non_editable/README.rst
```diff
@@ -0,0 +1,56 @@
...
+In odoo, by default, you can't add attachments to a record if the form view
+is set as not editable (``edit='false'`` in the form tag)
+
+This module adds a new form view attribute, ``attach=``, which lets you
+enable attachment uploads even in non-editable forms.
+
+Usage
+=====
+
+Add the ``attach=true`` attribute to the form view tag which of uneditable views
```
dreispt: Add the quotes: ``attach="true"``

dreispt (approved):
LGTM

hbrunn (approved):
people probably will be disappointed because this still doesn't allow attaching attachments to records which are actually not writable for the user, because this is how Odoo checks permissions. Maybe add a line about that to the readme?

attachment_attach_non_editable/README.rst
LeartS: Should I link the runbot to the 10.0 branch (which would work once merged) or to the pull request branch (which works as long as the PR is active, but not after)? I always had this doubt

LeartS:
Addressed the comments, @max3903 can you update your review?

max3903 (approved):

PR #136, opened 2017-05-02
[FIX] attachments_to_filesystem: Fix new API conversion

### LINKED ISSUE #135: PR #129 breaks attachments_to_filesystem
in this PR the code has been incorrectly moved to the new api:

in models/ir_attachment.py, line 83:

        ir_attachment = self.env['ir.attachment']
        attachment_ids = ir_attachment.search(
            [('db_datas', '!=', False)], limit=limit)

Since it's now the new api, the attachments are returned and no longer the attachment_ids.
The rest of the code still believes that a list of ids is returned and hence it breaks,
cf. line 89
            attachment_data = ir_attachment.read(
                [attachment_id], ['datas', 'res_model']

an quick fix could be to change [attachment_id] into [attachment_id.id] but it's probably cleaner to go the new api all the way (hence change 'attachment_ids' in line 84 into 'attachments' and skip the 'read' in line 89 since we have already the records in browse.

pedrobaeza:
New API search method returns a recordset, so we need to adapt code
to this condition. New API write method also requires a recordset.

Closes #135

### DIFF
--- a/attachments_to_filesystem/models/ir_attachment.py
+++ b/attachments_to_filesystem/models/ir_attachment.py
@@ -81,11 +81,11 @@ class IrAttachment(models.Model):
             self.env['ir.config_parameter'].get_param(
                 'attachments_to_filesystem.limit', '0')) or limit
         ir_attachment = self.env['ir.attachment']
-        attachment_ids = ir_attachment.search(
+        attachments = ir_attachment.search(
             [('db_datas', '!=', False)], limit=limit)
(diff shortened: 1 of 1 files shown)


pedrobaeza:
ping @luc-demeyer

hbrunn (approved):
thanks, I figured if I do it, I'd do the big refactoring, that's why I postponed... But yes, non-breaking old code is better than nonexisting new code

dreispt (approved):

sbidoul (approved):

PR #137, opened 2017-05-07
[ADD] document_wopi

hbrunn:
WOPI
====

This module was written to implement the [WOPI protocol](http://wopi.readthedocs.io/projects/wopirest) in order to allow online collaborative editing and viewing of office files. This works with [LibreOffice online](https://wiki.documentfoundation.org/Development/LibreOffice_Online) and probably also with Office 365.

Screenshot

Installation
============

To install this module, you need to:

- have a WOPI client available. For testing, use one of the [Collabora](https://hub.docker.com/r/collabora/code) or [Libreoffice](https://hub.docker.com/r/libreoffice/online/) docker images. The first works out of the box, but is rate limited, the second needs manual interaction, but is unmodified. The [Nextcloud documentation](https://nextcloud.com/collaboraonline) contains some information on how to get this running
- have your instance accessible on port 443 via your WOPI client and should have a valid SSL certificate
- use a dbfilter if necessary to have unauthenticated requests ending up in the database you want

Configuration
=============

To configure this module, you need to:

- fill in your WOPI client URL in the configuration wizard

Usage
=====

To use this module, you need to:

- on any supported attachment, click the edit button

Known issues / Roadmap
======================

- [X] implement actually modifying files
- [X] implement attachment locking in its own module for cooperation across different locking mechanisms
- [X] use WOPI discovery to determine which file types are viewable/editable
- [ ] whoever want to get their hands dirty with Microsoft should find our how to integrate this with them
- [ ] implement quasi permanent public shares a la google docs
- [X] finally write some base module that just defines a container for all other addons to put their attachment buttons


hbrunn:
we can't test this on runbot because a proper SSL configuration is mandatory.

@OCA/board @gurneyalex are there any plans to buy wildcard certificates for the runbots, or to support letsencrypt on runbot?

dreispt:
Let's encrypt sounds good to me, but I can't comment myself on the technical feasibility.

gurneyalex:
@hbrunn I don't think we can use letsencrypt on runbot, because they don't do wildcard certificate, and last time I checked there is a limit on the number of certificates they can issue per week.

Then there is the issue of wildcard certificates for *.runbot[12].odoo-community.org. I think we can have a wildcard certificate for *.odoo-community.org but not for the sub subdomain :-/ 

I'll ask around and report back.

hbrunn:
@gurneyalex @dreispt thanks for your input. I was not aware that multi level wildcard certificates on subdomains are a problem (cf https://www.ssl.com/faqs/can-i-create-a-subdomain-domain-com-wildcard-how-about-subdomain-com for example). What I was toying with on Therp's runbot was runbot_letsencrypt, that adds a button to allow a user to create a valid certificate for some currently running instance. Manually because doing this on every build would max out the rate limit very fast. How about doing this? If you agree, I'll put this on my list for community stuff to do.

And I'm a bit disappointed nobody commented on the serious coolness of this addon :-( I think this can be a huge thing.

pedrobaeza:
Hey, what a cool addon!! :stuck_out_tongue: 

No, seriously it is, but we need to assimilate how to change the rules of the ERP with this.

gurneyalex:
@hbrunn +1 for runbot-letsencrypt if we can manage this with the docker builds. 

re:  the coolness of the addon, I do agree this is very nice (although not in a part of Odoo I massively deploy). Parallel modification on most models in Odoo are still a bummer, though. In the best cases you get an error. In the not nice ones, you get corrupted data...

hbrunn:
@gurneyalex that's the beauty of this: When you edit a document simultaneously, you edit the *same* document (just like google docs, you'll also see the other users' modifications appear on your screen). What is missing here in the PoC is to lock the attachment so that users can't modify the attachment by different means. But concurrency of edits via libreoffice online is handled there. For the locking I need more time (=funding) to do this properly in a way that other modules (like `document_sftp`) can add locks via the same mechanism too.

gurneyalex (approved):

gurneyalex:
@hbrunn I totally get the point. My current issues are with 1 employee adding lines to a sale order while his colleague validates the same sale order, or 2 warehouse employees processing the same picking in parallel...

document_wopi/demo/ir_config_parameter.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
```
dreispt: noupdate="1"

document_wopi/demo/ir_config_parameter.xml
```diff
@@ -0,0 +1,9 @@
...
+        <record id="param_client" model="ir.config_parameter">
+            <field name="key">document_wopi.client</field>
+            <field name="value">https://docs.therp.nl</field>
```
dreispt: Are you sure you want to leave this URL here?

document_wopi/models/document_wopi_access_token.py
```diff
@@ -0,0 +1,88 @@
...
+            ))
+
+    @api.multi
+    def _compute_action_url(self):
+        # TODO: this needs to look up the value from the last call to
+        # /hosting/discovery. The current value probably is installation
+        # specific
+        for this in self:
+            this.action_url = self.env['document.wopi']._get_url(
+                'loleaflet/cda55f3/loleaflet.html'
```
dreispt: We should really know a little more about this value...

dreispt (commented):
Great feature, thanks!

document_wopi/models/document_wopi_access_token.py
hbrunn: yes, this is not production ready. WOPI does service discovery, we need to save those URLs and use them. This is specific for the current docker version of libreoffice online.

document_wopi/demo/ir_config_parameter.xml
hbrunn: there's not so much to worry about, WOPI clients only allow predefined domains (one of the reasons we need the SSL dance), so the only place to actually use this is from OCA's runbot and Therp's installation.
And in case the tests take too much of our resources, I can revoke access.

document_wopi/models/document_wopi.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @api.model
+    def _call(self, path, query=None, headers=None):
+        return requests.get(
+            self._get_url(path), params=query, headers=headers
+        )
+
+    @api.model
+    def get_access_token(self, attachment_id):
+        """Return an access token to pass to the WOPI client. Make sure the
```
lasley, line 30: `Return an access token ID` I think

document_wopi/models/document_wopi_access_token.py
```diff
@@ -0,0 +1,88 @@
...
+        return self.create({
+            'user_id': self.env.user.id,
+            'attachment_id': attachment.id,
+            'valid_until': fields.Datetime.to_string(
+                fields.Datetime.from_string(fields.Datetime.now()) +
+                timedelta(self.env['ir.config_parameter'].get_param(
+                    'document_wopi.token_validity', 36000,
+                ))
+            ),
+            # TODO: is this random enough or do we need some secret generator?
```
lasley: This is the way Odoo SA does it, but I don't think it's random enough personally. 

Probably best to not diverge though, so we can likely remove this comment? 

Fail2Ban4Lyfe, although this does remind me to put up my RFC for an internal 404 trap too.

lasley (commented):
Well this is neat

document_wopi/models/document_wopi_access_token.py
hbrunn: yes, I have the same doubts, hence the comment. I'll leave it there until there's a production ready version, so that we don't forget to do something better here. I'm curious to read your RFC, if that's about what I think it is, it's a great idea

lasley: https://github.com/OCA/server-tools/issues/835

hbrunn:
this is now testable on runbot (works only with exactly this build; If you have to rebuild, you'll have to change the base url to https), you can check Knowledge/Pages/OpenERP 6.1. Functional Demo

hbrunn:
@dreispt do you have time for a review?

attachment_action/__openerp__.py
```diff
@@ -0,0 +1,22 @@
...
+    "name": "Actions on attachments",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "category": "Hidden/Dependency",
+    "summary": "Enable/disable actions on attachments",
+    "depends": [
+        'web',
+        'knowledge',
+        'document',
```
dreispt, line 14: Are all these dependencies strictly necessary?

attachment_lock/README.rst
```diff
@@ -0,0 +1,72 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: https://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+==================
+Attachment locking
+==================
+
+This module was written to allow users to lock attachments for external editing.
```
dreispt, line 9: Sorry, its not clear to me the goal for the locking.
AFAICS the doc owner can create a lock, and one hour later it will be released.
The point is to avoid concurrent edits?
What about the case you want to lock the document forever (such as when it is a final version)?

dreispt (approved):
LGTM

attachment_action/__openerp__.py
hbrunn: thanks for the review Daniel. We indeed only need `web` for technical reasons, but `document` makes sense imho because otherwise, you don't see any attachments at all. And `knowledge` because I plan to add some of the modules depending on this to the document settings wizard

attachment_lock/README.rst
hbrunn: that's in the road map, will come. Libreoffice Online will request a new lock btw, so the hour is basically the TTL. The purpose here is that we need a way to be sure nobody else edits an attachment when we edit it with external means. And I made this a module instead of hacking it into `document_wopi` itself because `document_sftp` needs just the same for proper write access.

This one will be going slow btw, because it's a pet project of mine, no day job work.

PR #129, opened 2017-04-02
[FIX] ignore exceptions when moving attachment

hbrunn:

PR #138, opened 2017-05-08
fix attachments_to_filesystem

luc-demeyer:
correction for https://github.com/OCA/knowledge/issues/135

### DIFF
--- a/attachments_to_filesystem/__openerp__.py
+++ b/attachments_to_filesystem/__openerp__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 {
     "name": "Move existing attachments to filesystem",
-    "version": "8.0.1.0.0",
+    "version": "8.0.1.0.1",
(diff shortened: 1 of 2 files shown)


pedrobaeza:
What about Travis?

lasley:
And maybe a unit test to cover this, seeing as it already bit us in the ass once?

luc-demeyer:
The travis failure is unrelated to this fix. 
Adding a unit test would of course be a lot better but the question is who has time to do it.
In the meanwhile we should merge this fix or fully revert PR #129 because now we have a broken module which is worse.

pedrobaeza (approved):
Well, it was broken before that PR also. Travis is unrelated indeed.

lasley (approved):

hbrunn:
tests will be helpful anyways, I only merged this without them because it fixes a problem

PR #146, opened 2017-09-25
8.0 document definition

gfcapalbo:
Addd new small module to add a dictionary style module to knowledge, with aplphabetical grouping.


document_definition/README.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* Giovanni Francesco Capalbo <giovanni@therp.nl>
```
thomaspaulb, line 29: Isnt this gcapalbo?

document_definition/security/ir.model.access.csv
```diff
@@ -0,0 +1,3 @@
+"id","name","model_id:id","group_id:id","perm_read","perm_write","perm_create","perm_unlink"
+document_definition_all,document.definition,model_document_definition,,1,1,1,1
+
```
thomaspaulb: empty line

document_definition/models/document_definition.py
```diff
@@ -0,0 +1,39 @@
...
+
+class DocumentDefinition(models.Model):
+    _name = 'document.definition'
+
+    term = fields.Char(help='Term to define', required=True)
+    definition = fields.Text(help='Definition of term', required=True)
+
+    # stored compute field used for making the letter grouping
+    # need to be stored in order to use grouping
+    first_letter = fields.Char()
```
thomaspaulb, line 15: If it is to be a stored computed field, any reason why you aren't just using `compute=` and `stored=`?

document_definition/models/document_definition.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def _get_first_letter(self, text):
+        # if somehow the user starts with a blank space we trim it
```
thomaspaulb: Also with more than one blank space,

document_definition/README.rst
```diff
@@ -0,0 +1,54 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+===================
+document_definition
+===================
+
+Installs a new  document_definition model for dictionary-style word archive and
+definition. Searchable in the Knowledge menu.
```
thomaspaulb: Searchable on which view? Can you describe the steps a user would do to use the functionality of this module as per [here](https://github.com/OCA/maintainer-tools/blob/master/template/module/README.rst#usage)

document_definition/models/document_definition.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, fields, models
+
+
+class DocumentDefinition(models.Model):
+    _name = 'document.definition'
+
+    term = fields.Char(help='Term to define', required=True)
```
thomaspaulb: I would say just 'Term' or 'Dictionary term' here, otherwise they refer to each other

document_definition/README.rst
```diff
@@ -0,0 +1,54 @@
...
+
+Known issues / Roadmap
+======================
+
+* A search function per term and per definition, obviously.
+* Dictionary Style Alphabetical grouping view
```
thomaspaulb: Something I am also missing (but maybe I have overlooked it, or am just not understanding completely) is a view or field that allows the user to link the terms to a knowledge base document?

thomaspaulb (changes requested):
One thing am not getting completely is: do users add dictionary terms by hand?

document_definition/README.rst
```diff
@@ -0,0 +1,54 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+===================
+document_definition
```
hbrunn: use a pretty name

hbrunn (changes requested):
this PR contains js and css that don't do anything.

But what i ask myself: Why is this not a different view on `document.page`? This is the place where we're supposed to save knowlegde anyways, and a definition seems a very knowledgy-thing to me.

document_definition/README.rst
gfcapalbo: in the view "Definitions" when expanding "advanced search" you will see an option "alphabetical" where the terms are grouped by their first letter, like a dictionary.

document_definition/README.rst
gfcapalbo: github handle gfcapalbo, shouldn't the proper name be here?

document_definition/models/document_definition.py
gfcapalbo: yes agreed

document_definition/models/document_definition.py
gfcapalbo: computed fields are not allowed in groupings, as we use it. Only database persistent data is allowed.
Logically I thought also that a stored computed field would had met this requirement, but odoo refuses to accept any computed field as a grouping function.

gfcapalbo:
@thomaspaulb @hbrunn  closing this a better datamodel has been devised. will show up with the new version

PR #155, opened 2018-02-28
[IMP][9.0] Change Requests and workflow improvements on documents

ivantodorovich:
This PR introduces these changes to document_page and document_page_approval.
It's intended to improve document managing workflow, specially in the use-case of [OCA/management-system](https://github.com/OCA/management-system)

### User experience changes
- Added alerts to let the user know what's going to happen if he edits the document
- It's now possible to add comments (`summary`) to each review.
- Change requests (history) have a more complete workflow
- Added a `template` field to be used as a template for categories.
- `diff` is now css-styled and shown on email notifications
- `diff` can also be viewed from the history
- Improved several views

### Security
- Introduced new user groups for better security.
- Approvers need to be members of Approver's group.

### Backend changes
- Cleaned up code according to OCA Guidelines.
- `display_content` is deprecated. Replaced with just `content` as a computed field.
- History is now deleted `oncascade` with pages.
- `diff` is now a computed field, not just a wizard.
- `diff` is now a little bit more readable by hacking in some new lines.
- `name_get` for `document.page.history`
- Pages that don't require approval still go through the workflow but gets auto-approved
- On install, `document_page_approval` sets all previous history to `approved`
- On uninstall, it deletes all non-approved history (change requests)

### Screenshots

image

image

image


ivantodorovich:
@max3903 I've noticed you've migrated this modules to 10.0.
Would you please take a look a this? I don't mind migrating this changes to 10.0 if it gets approved.

max3903 (approved):
:+1: 

1 pull request for each module would have been nice and easier to review the code.

ivantodorovich:
Yes, you are right. Sorry about that.

document_page/models/document_page.py
```diff
@@ -99,60 +117,58 @@ class DocumentPage(models.Model):
...
         if index:
             r += "<ul>" + "".join(index) + "</ul>"
         return r
 
-    def _get_display_content(self):
-        """Return the content of a document."""
-        for page in self:
-            if page.type == "category":
-                display_content = self._get_page_index(page, link=False)
+    @api.multi
```
zaoral, line 136: Use api.depends for this compute field.

document_page/models/document_page.py
```diff
@@ -99,60 +117,58 @@ class DocumentPage(models.Model):
...
+    def _compute_display_content(self):
+        # @deprecated, simply use content
+        for rec in self:
+            rec.display_content = rec.content
+
+    @api.multi
+    def _compute_content(self):
```
zaoral: same here

document_page/models/document_page_history.py
```diff
@@ -33,22 +30,38 @@ class DocumentPageHistory(models.Model):
...
 
-    page_id = fields.Many2one('document.page', 'Page')
+    page_id = fields.Many2one('document.page', 'Page', ondelete='cascade')
     summary = fields.Char('Summary', index=True)
     content = fields.Text("Content")
-    create_date = fields.Datetime("Date")
-    create_uid = fields.Many2one('res.users', "Modified By")
+    diff = fields.Text(compute='_compute_diff')
+
+    @api.multi
```
zaoral, line 37: same here

document_page/models/document_page.py
ivantodorovich: Since it's not stored, I think it's not actually needed.

ivantodorovich: I've just checked and indeed I've already added depends here: https://github.com/OCA/knowledge/pull/155/commits/dbeec6ef0f76f629bdf2d96c469c2fa4d25f20c7

ivantodorovich: @zaoral where you able to look at this?

zaoral: @ivantodorovich yes thank you for apply the api.depends. LGTM

zaoral (approved):

ivantodorovich:
Is this ready to merge or do we need more reviews?

document_page/models/document_page_history.py
max3903: @ivantodorovich I think we still need an api.depends here, no?

ivantodorovich: Since it's not stored, it's actually not needed.
Function gets called every time you access the field.

ivantodorovich: Just in case, just added it.

PR #165, opened 2018-04-28
document_url: Fix inclusion of unknown field 'user_id'

ernestotejeda:
Remove 'user_id' key in a dictionary prepared to
create an 'ir.attachment' object, because it's
an unknown field for that model

cc @Tecnativa


pedrobaeza:
Please change module version number according https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#version-numbers

rafaelbn:
cc @Tecnativa

PR #166, opened 2018-04-28
document_url: Fix inclusion of unknown field 'user_id'

ernestotejeda:
Remove 'user_id' key in a dictionary prepared to
create an 'ir.attachment' object, because it's
an unknown field for that model

cc @Tecnativa


pedrobaeza:
This one is still not migrated. You should point to this problem in the PR #111

ernestotejeda:
In the PR  #111 this problem is resolved

rafaelbn:
cc @Tecnativa

PR #172, opened 2018-06-20
[IMP] knowledge: add missing menus

cubells:
Add menus removed in migration.

cc @Tecnativa

### DIFF
--- a/knowledge/__manifest__.py
+++ b/knowledge/__manifest__.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 {
     "name": "Knowledge Management System",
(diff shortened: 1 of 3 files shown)


knowledge/i18n/knowledge.pot
```diff
@@ -1,133 +0,0 @@
-# Translation of Odoo Server.
```
pedrobaeza: Please don't remove this. It's now added by the new Weblate bot because it's needed.

cubells:
@pedrobaeza done.

pedrobaeza (approved):
Working as expected. Fast-tracking when CIs finish

MiquelRForgeFlow (approved):

PR #176, opened 2018-07-12
[MIG] document_page_partner_id

hbrunn:


max3903:
Hi @hbrunn, I don't know the details of your use case, but what about using the chatter and add the partner as a follower?

hbrunn:
We use document pages to document internal processes of our customers. And because we enforce access rules so that only employees who work for a specific customer's projects have access to the information about this customer (was a good thing to do before the GDPR, is mandatory now depending on the nature of the data), I prefer to have an explicit field for this instead of burying it in the followers. Our private module on top of this makes the field mandatory, that's quite impossible to implement in a non-convoluted way with a many2many.

I understand it's a tiny module, so if you prefer to leave this out, fine with me. But we do have customers who use this too, so I figure it's useful.

max3903 (approved):

max3903:
@hbrunn Thanks for the explanation.

:+1:

ivantodorovich (approved):

pedrobaeza:
Isn't better to add the field to `document_page`? This is very simple and you can leave not required for having the "previous" behavior.

ivantodorovich:
> Isn't better to add the field to `document_page`? This is very simple and you can leave not required for having the "previous" behavior.

@pedrobaeza I don't think it makes sense, giving the general purpose of the module.

There are many usecases similar to this one that I can think of. 
For instance, it could be possible to link document_pages to projects, manufacturing orders, sales orders... but all of those are very specific to the customer requirement. I don't think they should be on `document_page`, but rather in separate modules like this.

Maybe the question is: should there be a module for this? I mean, it's a very specific usecase and it's something you can achieve with customs fields and views anyways.

PR #182, opened 2018-08-24
[FIX] remove the set and get values in res.config

nicomacr:
This methods are not necessary for configs of type module_* and it is dangerous because if parameter does not exist and module is already install, when you click on apply of the config then it uninstall all this modules and the ones that depends on them.


zaoral (approved):
+1
Excellent @nicomacr 

Indeed. ´module_x´ fields are handle directly by odoo, This is dangerous and unnecessary

Thanks for this PR

pedrobaeza (approved):

PR #188, opened 2018-09-15
[MIG] document_page_multi_company

ivantodorovich:
Added this feature from the previous `document_page_multi_company` directly in `document_page`.

### DIFF
--- a/document_page/models/document_page_history.py
+++ b/document_page/models/document_page_history.py
@@ -24,6 +24,8 @@ class DocumentPageHistory(models.Model):
         'Company',
         help='If set, page is accessible only from this company',
         related='page_id.company_id',
+        store=True,
+        index=True,
         readonly=True,
     )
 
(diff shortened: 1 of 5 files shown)


document_page/models/document_page_history.py
```diff
@@ -19,6 +19,14 @@ class DocumentPageHistory(models.Model):
     content = fields.Text()
     diff = fields.Text(compute='_compute_diff')
 
+    company_id = fields.Many2one(
```
pedrobaeza, line 22: I think this one should be related to the page. It's a nonsense to have a revision under one company and the main page under other, don't you think?

ivantodorovich: It is related to the page

pedrobaeza: Sorry, I didn't see it! Any way, it would be good to be stored for avoiding penalty when record rule is evaluated.

ivantodorovich: I agree! Added index aswell

pedrobaeza (approved):

max3903 (approved):

PR #189, opened 2018-09-15
Add icon master file

ivantodorovich:
As suggested in https://github.com/OCA/management-system/pull/233#issuecomment-421599435

Adding the icon SVG master file here too


pedrobaeza (approved):
Fast-tracking this

PR #196, opened 2018-12-11
[11.0]Remove unrequired style color

SalahAdDin:
It shows the element in white color over a white background
deepin-screen-recorder_select area_20181211164619
.


pedrobaeza:
You should try without that backend theme.

SalahAdDin:
@pedrobaeza It works fine without theme, you are right, i'm sorry.

pedrobaeza:
I don't know why so much people like this backend theme. I personally hate it, as it doesn't add anything than web_responsive doesn't do, except styles, which are less ergonomic.

PR #201, opened 2019-01-17
py3 support

marianoadangelo:


pedrobaeza:
Can you explain? It was working already in Python 3

ivantodorovich:
Indeed this is not needed, in fact it's breaking the code because `parse` is not defined

PR #205, opened 2019-02-28
[knowledge][11.0.3.0.0] - add security for access to ir.attachment.

JordiBForgeFlow:
Otherwise the module grants too wide permissions for all users.

Users with permission *Central access to Documents* can access in
*Knowledge/Documents* to all the documents attached to records of any model
for which they have read permission.

To set up this module, you need to go to:

* Knowledge / Configuration / Settings

From this menu you'll have a central access to install the apps that belong
to Knowledge.

* Check *Attachments List and Document Indexation* if you want to install the
  module that allows users to attach documents to any model.
* Check *Manage attachments centrally* if you want all users to be able to
  access to the all attachments to which they have read permissions, from the
  menu *Knowledge / Documents*

If you want to grant Central Access to Documents only to some users:

#. Go to *Settings/Activate the developer mode*. Only a user with
  *Administration / Settings* permissions can do that.

#. Go to *Settings / Users & Companies / Users* and set the checkbox
   *Central access to Documents* to the selected users.

image


LoisRForgeFlow (approved):
Makes sense :+1:

mreficent:
Could you rebase, please?

MiquelRForgeFlow (approved):
LGTM

PR #207, opened 2019-03-12
[12.0][MIG] knowledge

MiquelRForgeFlow:
Standard migration.


knowledge/models/ir_attachment.py
```diff
@@ -1,6 +1,5 @@
 # Copyright (C) 2004-2009 Tiny SPRL (<http://tiny.be>).
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
```
pedrobaeza: No need to do this change. I even prefer the other way

knowledge/models/res_config.py
```diff
@@ -1,11 +1,9 @@
 # Copyright (C) 2004-2009 Tiny SPRL (<http://tiny.be>).
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
```
pedrobaeza: No need of this

pedrobaeza (approved):

knowledge/models/ir_attachment.py
MiquelRForgeFlow: why do you prefer the white line?

pedrobaeza: For readability. We have to avoid it inside methods, but here helps to better focus the code line, which doesn't happen when stick to the comment lines with the usual IDE highlighting. The point is to not touch this in a migration and keep the style decided by the original contributor, for reducing the diff

JordiBForgeFlow (approved):

PR #213, opened 2019-05-22
[11.0][FIX] document_page: use document page view as default instead of category one.

LoisRForgeFlow:
This was specially annoying when accessing a document with planned activities from the topbar activities dropdown.

@Eficent


JordiBForgeFlow (approved):

JordiBForgeFlow:
Fast tracking this one.

PR #215, opened 2019-05-31
[FIX] document_page_approval: Modify in records rules

nicomacr:
Add the the group eployee to the rule "Change Request Global" to allow to the user with the rule "Change Request Approver" to create pages

### DIFF
--- a/document_page_approval/security/document_page_security.xml
+++ b/document_page_approval/security/document_page_security.xml
@@ -14,6 +14,7 @@
     <record model="ir.rule" id="rule_change_request_global">
         <field name="name">Change Request Global</field>
         <field name="model_id" ref="model_document_page_history"/>
+        <field name="groups" eval="[(6, 0, [ref('knowledge.group_document_user')])]"/>
         <field name="domain_force">['|',('state','=','approved'),('create_uid','=',user.id)]</field>
         <field name="perm_read" eval="True"/>
         <field name="perm_write" eval="True"/>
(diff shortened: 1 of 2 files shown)


max3903:
@nicomacr If you had a group to the rule, the rule is not global anymore and only applies to that group. This is a major change of behavior.

This should be an option in the settings.

document_page_approval/security/document_page_security.xml
```diff
@@ -14,6 +14,7 @@
     <record model="ir.rule" id="rule_change_request_global">
         <field name="name">Change Request Global</field>
         <field name="model_id" ref="model_document_page_history"/>
+        <field name="groups" eval="[(6, 0, [ref('base.group_user')])]"/>
```
ivantodorovich: It makes more sense to use `knowledge.group_document_user` because that's the minimum level of the knowledge app.

ivantodorovich:
I can confirm there is a bug there, that rule should definitely not be global.

To reproduce the bug on runbot: 
1. set demo user to Knowledge / Approver
2. activate changes on the document category
3. create a change request using another user (ie: admin)
4. try to validate the change request using demo

The change request doesn't show because of that global rule, so it's impossible to approve.


Tested this on runbot and its fixed now :+1:
I would still change the group, though

ivantodorovich (changes requested):

ivantodorovich (approved):

max3903 (approved):

PR #217, opened 2019-06-05
[FIX] lint errors in README files

gurneyalex:
and add missing README.rst files (copied from 9.0 branch)


pedrobaeza (approved):
Thanks for this. Fast-tracking it

PR #218, opened 2019-07-11
[12.0] document_page:  fix links in category index

simahawk:
ATM  links are broken. They just use `#id`. Completely unusable.

The 1st commit fixes it by adding the model: the link now works.

Then, w/out an action, the link works but the context of the app is lost.
The 2nd commit tries to improve this by using the menu assigned to the specific page or it looks up in the page hierarchy until it finds one.
This implies that if you want a nice behavior you have to assign menu_id to all the categories when you set them up.

BTW I don't understand why the menu has been made not editable in some views.

### DIFF
--- a/document_page/models/document_page.py
+++ b/document_page/models/document_page.py
@@ -101,6 +101,17 @@ class DocumentPage(models.Model):
         index=True,
         ondelete='cascade',
     )
+    backend_url = fields.Char(
+        string='Backend URL',
+        help='Use it to link resources univocally',
+        compute='_compute_backend_url',
+    )
+
+    @api.depends()
+    def _compute_backend_url(self):
+        tmpl = '/web#id={}&model=document.page&view_type=form'
+        for rec in self:
+            rec.backend_url = tmpl.format(rec.id)
 
     @api.constrains('parent_id')
     def _check_parent_id(self):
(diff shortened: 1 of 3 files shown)


max3903 (approved):

pedrobaeza (commented):
Isn't a general method already that do this? @etobella I think you used it for your module qr_action or similar.

etobella:
Well, `quick_document_access` uses the standard function `get_formview_id` that returns the standard formview of a record. 
I was checking and you should probably use `get_formview_action` to obtain the expected view and edit `get_formview_id` on `document.page` in order to obtain the expected form. You can check how it was done on `account.invoice`

simahawk:
@etobella @pedrobaeza thanks for your feedback. I might be wrong but we are not talking about the same thing.

One thing is to retrieve the form and _an_ action and one thing is retrieve the _right_ action.
As you know, if you don't use the right action you lose the context (breadcrumb, menu, etc)

We don't have any issue w/ the form: it works and it works also because you pick - now - the right action (which could lead to different forms if configured properly).

`get_formview_action` is a default scaffolding for the action dict, you can do it even manually, it has no clue of where you are in the app.

I hope I'm clear :smile:

vrenaville (approved):

i-vyshnevska (approved):

pedrobaeza:
Are you sure the URL you are returning is aware of the breadcrumb? There's no annotation in the URL of the breadcrumb, so I think the only difference for that is the target of the action.

simahawk:
sure is the action, but calling `get_formview_action` won't give you the right action. 
That's why I'm using it. Retrieving it from the menu (that we can already customize) seems the right thing to do because you will always use either the one from the category or the one of the custom menu.

pedrobaeza:
Ok, if you say so, I haven't dug enough

PR #221, opened 2019-09-20
[12.0][MIG] attachment_preview

tarteo:


astirpe (approved):
Tested in Runbot. Works fine!
Just the buttons to open the preview are very small. Would be nice to have only one button that opens a dropdown containing a more readable list of actions. But that's something that could be done in the future as enhancement.

jcdrubay:
Tested in local, worked well.
But I wonder why it is in this repo.... Shouldn't it be in web or server-ux ?

attachment_preview/model/ir_attachment.py
```diff
@@ -0,0 +1,68 @@
...
+                    import magic
+                    ms = magic.open(
+                        hasattr(magic, 'MAGIC_MIME_TYPE')
+                        and magic.MAGIC_MIME_TYPE or magic.MAGIC_MIME)
+                    ms.load()
+                    mimetype = ms.buffer(
+                        base64.b64decode(this[binary_field]))
+                except ImportError:
+                    (mimetype, encoding) = mimetypes.guess_type(
+                        'data:;base64,' + this[binary_field], strict=False)
```
yucer: Is this method of guessing the mime type still optimal ?

We have now the method [guess_mimetype](https://github.com/odoo/odoo/blob/12.0/odoo/tools/mimetypes.py#L135) in tools.

It fails when `python-magic` is not installed. 

In the section after `ImportError`, the string literal: `'data:;base64,'` is now _unicode_ in **python3** and fails with one **unicode** error instead of trying another guess as intended by the code.

yucer: I mean The code is supposed to have an alternate method to that `magic`. That code is not being included in the coverage and it is failing.

yucer (changes requested):
The alternate code to python-magic method of guessing the mime-type doesn't work in python3.

It fails also in **v11**, but it might be a good opportunity to fix it.

attachment_preview/model/ir_attachment.py
hhgabelgaard: Tested in runbot. Works very well.
Any plans to let the split screen functionality being auto-triggered by the 
`<div class="o_attachment_preview" attrs="{'invisible': ['|',('type', '!=', 'in_invoice'),('state', '!=', 'draft')]}"/>`
 found in Odoo Core in various places (e.g. on Vendor bills)?

pedrobaeza:
Do you know how this interacts with `web_responsive` with side chatter?

attachment_preview/model/ir_attachment.py
yucer: @hhgabelgaard:   I guess you wanted to make a functional review and ended commented my review.

I am not talking about the functional part of the module here. I bet it is a good work done there. What I am pointing out here is not something that you can see in the user interface. it is about the way to handle dependencies.

Runbot works ok, because the new [code](https://github.com/OCA/knowledge/blob/063c58e25a46589f39d972fea6350328e4f40a5f/.travis.yml#L33) tweaks the **runbot** installation process to install `python-magic`.

When python-magic is not installed the code fails. And python magic seems to be marked as a dependency of Odoo in [v12](https://github.com/odoo/odoo/blob/12.0/requirements.txt), nor from the **addon** dependencies.

yucer: No answer for this ? it is still giving error in our tests instances.

hhgabelgaard:
@pedrobaeza It kind of work with web_responsive and side chatter:

image

But I think it would work better if it moved the chatter to the bottom on preview in side panel (split view)

pedrobaeza:
Yeah, it's not the best option indeed. Can it be possible to make compatible both chatter positions? Also enterprise have a module for automatically put that preview if it's the main attachment. Can it be interesting to be added to this module?

pedrobaeza:
We have implemented in https://github.com/OCA/web/pull/1467 the side document viewer if you want to check.

MiquelRForgeFlow:
@pedrobaeza With https://github.com/OCA/web/pull/1467 already merged, is this PR already needed?

pedrobaeza:
Well, they decide.

tarteo:
@MiquelRForgeFlow I think this should be retained as some customers don't like the responsive layout (for some readon). The document preview in web_responsive is only available if the chatter position is to the right. I see no problem in a little bit redundancy here.

MiquelRForgeFlow (approved):
Good!

alexey-pelykh:
Isn't side document viewer only available in enterprise version (for web_responsive to use it)?

etobella (approved):
LGTM

etobella:
> Isn't side document viewer only available in enterprise version (for web_responsive to use it)?

It is available without enterprise

etobella:
Well, I would prefer to have both modules integrated, but, I understand both ideas and IMO the best solution right now is to maintain both modules. I understand this module could be necessary if web_responsive is not installed
/ocabot merge nobump

etobella:
Ocabot is not runing properly... I will try again, let's see if it is enough:
/ocabot merge nobump

PR #225, opened 2019-11-04
[MOD]document_sftp_root_by_model.py

MarcoFolco:
Changed _list_folder method to show "<res_id> - <name>" instead of only <res_id> in the ftp.

I think this change will make the search through the attachments more comfortable for the user.

### DIFF
--- a/document_sftp/models/document_sftp_root_by_model.py
+++ b/document_sftp/models/document_sftp_root_by_model.py
@@ -2,7 +2,7 @@
 # © 2016 Therp BV <http://therp.nl>
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 import os
-from openerp import api, models
+from openerp import api, models, _
 try:
     from paramiko import SFTP_NO_SUCH_FILE, SFTP_PERMISSION_DENIED
 except ImportError:   # pragma: no cover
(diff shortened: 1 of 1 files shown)


document_sftp/models/document_sftp_root_by_model.py
```diff
@@ -64,11 +64,18 @@ def _list_folder(self, path):
...
-                    seen.add(attachment.res_id)
-                    result.append(self._directory(str(attachment.res_id)))
+                    res_id = attachment.res_id
+                    res_model = attachment.res_model
+                    seen.add(res_id)
+                    res = self.env[res_model].browse(res_id)
+                    if res.name:
+                        display_name = str(res_id) + ' - ' + res.name
+                    else:
+                        display_name = str(res_id) + ' - No Name'
```
hbrunn: this should be translated

document_sftp/models/document_sftp_root_by_model.py
```diff
@@ -64,11 +64,18 @@ def _list_folder(self, path):
             ], order='res_id asc'):
                 # TODO: better lump ids together in steps of 100 or something?
                 if attachment.res_id not in seen:
-                    seen.add(attachment.res_id)
-                    result.append(self._directory(str(attachment.res_id)))
+                    res_id = attachment.res_id
+                    res_model = attachment.res_model
+                    seen.add(res_id)
+                    res = self.env[res_model].browse(res_id)
+                    if res.name:
```
hbrunn: this will break for all models without a name. Use `display_name`

hbrunn (changes requested):
my reasoning for not adding any info about the records is that this can change. But with you patch, we never cate for what comes after the ID, so I think that's fine too

PR #229, opened 2019-11-27
[ADD] document_page_reference

etobella:
This module allows to add a reference name on documents and simplifies the link
between document pages.

Usage
=====

When editing a document page add elements like ${XXX} where XXX is the reference
of another page. Now, when viewing the document, it will link directly to the page.
Also, the name will be parsed as the display name.


etobella:
Travis is failing because of #230

document_page_reference/__manifest__.py
```diff
@@ -0,0 +1,21 @@
...
+    'author': 'Creu Blanca,Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/knowledge',
+    'depends': [
+        'document_page',
+    ],
+    'data': [
+        'views/document_page.xml',
+        'views/report_document_page.xml',
+    ],
+    'demo': [
```
jarroyomorales: Remove if empty

document_page_reference/models/document_page.py
```diff
@@ -0,0 +1,88 @@
...
+            return document
+        return False
+
+    def get_reference(self, code):
+        element = self._get_document(code)
+        if not element:
+            return code
+        if self.env.context.get('raw_reference', False):
+            return element.display_name
+        return '<a href="%s">%s</a>' % (
```
jarroyomorales: Here you should probably use some regex or external libraries to remove html tags and prevent html injection

jarroyomorales (commented):
LGTM! Tested on runbot and it is an awesome feature!

document_page_reference/models/document_page.py
```diff
@@ -0,0 +1,88 @@
...
+
+    reference = fields.Char()
+    content_parsed = fields.Html(compute='_compute_content_parsed')
+
+    @api.depends('history_head')
+    def _compute_content_parsed(self):
+        for record in self:
+            record.content_parsed = record.get_content()
+
+    @api.constrains('reference')
```
jarroyomorales, line 59: It is easier to write it like this
    _sql_constraints = [
        ('reference_uniq', 'unique (reference)', "Reference must be unique !"),
    ]

LoisRForgeFlow:
This is really cool. Just one usability issue I noticed: when you navigate using this reference links you lose the breadcrumb and the menus, see:

no-breadcrums

document_page_reference/models/document_page.py
etobella: Well, reference could be not defined, and also, we must check the format, so I think it is better to maintain the current constrain

document_page_reference/models/document_page.py
etobella: Comments attended

document_page_reference/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2019 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Document Page Reference',
+    'summary': """
+        Include references on document pages""",
+    'version': '11.0.1.0.0',
+    'license': 'AGPL-3',
```
LoisRForgeFlow, line 9: I would add here yourself as maintainer and a alpha development status

LoisRForgeFlow (commented):
Now looks better, see a suggestion inline

LoisRForgeFlow (approved):
LGTM :+1:

etobella:
Now breadcrumbs is working as expected
:green_apple: :green_apple: :green_apple: :green_apple: 
@jarroyomorales @lreficent

jarroyomorales (approved):
Nice!

LoisRForgeFlow (approved):
Great! :clap:

AdriaGForgeFlow (approved):
Nice work :+1:

JordiBForgeFlow (approved):
:+1: Works like a charm!!

PR #231, opened 2019-11-29
[11.0] [ADD] document_page_group

etobella:
This module allows to define groups on a category.


document_page_group/models/document_page.py
```diff
@@ -0,0 +1,32 @@
...
+        relation='document_page_direct_group',
+        column1='document_page_id',
+        column2='group_id',
+        compute='_compute_group_ids'
+    )
+    direct_group_ids = fields.Many2many(
+        'res.groups',
+        relation='document_page_group',
+        column1='document_page_id',
+        column2='group_id',
```
LoisRForgeFlow, line 25: I would just a small help for the user, as the names themselves can not be enough

LoisRForgeFlow (approved):
Tested on runbot and code review, LGTM!

document_page_group/models/document_page.py
etobella: Well groups is hidden, but we could define something for direct_group_ids

LoisRForgeFlow: It is not hidden, see in runbot with demo user:
image

etobella: Ups, my mistake then :wink:

LoisRForgeFlow: Now it looks better form the user perspective :+1:

etobella:
@jarroyomorales

jarroyomorales (approved):
Tested and works as expected! :+1:

PR #244, opened 2020-03-18
[13.0][MIG] document_page_approval

flachica:
@NachoAlesLopez @ivantodorovich Need review


flachica:
The error is related to #245

flachica:
Testing dependencies. Need help for solve it

.pre-commit-config.yaml
```diff
@@ -1,99 +1,68 @@
...
+  rev: v4.3.21
+  hooks:
+  - id: isort
+    name: isort except __init__.py
+    exclude: /__init__\.py$
+- repo: https://github.com/pre-commit/mirrors-eslint
+  rev: v6.5.1
+  hooks:
+  - id: eslint
+    verbose: true
```
ivantodorovich, line 68: This file shouldn't be here :sweat_smile:

ivantodorovich (commented):
I think tests are failing because of the `_compute_is_approval_required` method.
It's giving False when it should be True. Maybe because of changes in the ORM re: computed methods?
It could be this : https://github.com/flachica/knowledge/blob/13.0-mig-document_page_approval/document_page_approval/models/document_page.py#L77 ?

Not sure, though

.pre-commit-config.yaml
flachica: No, the problem is than in #245 I correct a method dependency of this module

But I need to close this PR. I'm still having problems squashing commits. New PR is comming

flachica: New PR #246

PR #245, opened 2020-03-19
Fix document_page: Required impossible. Protected method needed

flachica:
Required impossible for type category in document_page
Protected method needed for dependencies

PR #246, opened 2020-03-23
[13.0][MIG] document_page_approval: Migration to 13.0

flachica:


ivantodorovich:
Thanks @flachica !

Can you follow the process in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0#technical-method-to-migrate-a-module-from-120-to-130-branch using the format-patch command to keep the commit history?

flachica:
@ivantodorovich Hope it is ok now

ivantodorovich:
Great! commit history is ok :)
Tests on Travis are failing though, can you check?

flachica:
@ivantodorovich  To pass all test I need the PR #245 has be merged previosly. As you can see [here](https://github.com/OCA/knowledge/blob/13.0/document_page/models/document_page.py#L145), _create_history protected method has desapear but in my [fix](https://github.com/flachica/knowledge/blob/13.0-document_page-fix-createhistory/document_page/models/document_page.py#L148) is added. With my fix all test run correctly

etobella:
@flachica Rebase it in order to fix travis please :wink:

flachica:
Rebased @etobella , but I dont understand why codecov is lower

etobella:
Probably related to other modules

document_page_approval/tests/test_document_page_approval.py
```diff
@@ -46,7 +44,7 @@ def test_change_request_approve(self):
 
         # It should automatically be in 'to approve' state
         self.assertEqual(chreq.state, "to approve")
-        self.assertNotEqual(chreq.content, page.content)
+        self.assertEqual(chreq.content, page.content)
```
etobella: Why are you changing the test?

document_page_approval/views/document_page_approval.xml
```diff
@@ -164,7 +164,7 @@
                 <field name="has_changes_pending_approval" invisible="1" />
                 <field name="user_has_drafts" invisible="1" />
             </sheet>
-            <button name="toggle_active" position="after">
+            <widget name="web_ribbon" position="after">
```
etobella: Why on web ribbon? Can you use the button_box?

etobella (changes requested):
I don't understand why do you need to change the tests? It seems the behavior has changed...

document_page_approval/views/document_page_approval.xml
flachica: I thought that [here](https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0) it was commented that the new paradigm is to use the web_ribbon

document_page_approval/tests/test_document_page_approval.py
flachica: Excuse me, I didn't remember that I had done that. In Odoo 13.0 the calculated fields have been improved and now we have to force the calculation. New rebase in progress

document_page_approval/views/document_page_approval.xml
etobella: web_ribbon is used to show that the element is archived, toggle_button needs to be used for buttons, so change 
`<widget name="web_ribbon" position="after">`
for
`<xpath expr="//div[@name='button_box']" position="inside">`

document_page_approval/tests/test_document_page_approval.py
etobella: Isn't `flush` enough? I mean, do `chreq.flush()` before.
`self.env.add_to_compute(page._fields["content"], page)` shouldn't be necessary

document_page_approval/views/document_page_approval.xml
flachica: Oh! Sorry! I understand. They are issues that have nothing to do with each other. Change made.

document_page_approval/tests/test_document_page_approval.py
flachica: Flush is not enough. I have tried previously but it didn't work

etobella:
@flachica Can you replace the computes on the tests by `flush` as explained on the comment?

flachica:
> @flachica Can you replace the computes on the tests by `flush` as explained on the comment?


Do you mean to replace [this](https://github.com/flachica/knowledge/blob/bb78963e012bf56551d9dc6f9de9bf691a180981/document_page_approval/tests/test_document_page_approval.py#L49) line with `chreq.flush()` and `page.flush()`?

etobella:
I have tested it locally, you are right, `flush` is not enough, you need to use the following: `page.refresh()`
The problem is that cache is not recomputed.

Also, can you rebase with the last version in order to install properly on a local version?

etobella:
runbot is failing because the rebase is required

flachica:
> I have tested it locally, you are right, `flush` is not enough, you need to use the following: `page.refresh()`
> The problem is that cache is not recomputed.
> 
> Also, can you rebase with the last version in order to install properly on a local version?

This is great. Solution more clear. Rebase in progress

document_page_approval/hooks.py
```diff
@@ -0,0 +1,25 @@
...
+            approved_date=create_date
+        WHERE state IS NULL OR state = 'draft'
+    """
+    )
+
+
+def uninstall_hook(cr, registry):  # pragma: no cover
+    # Remove unapproved pages
+    _logger.info("Deleting unapproved Change Requests.")
+    cr.execute("DELETE FROM document_page_history " "WHERE state != 'approved'")
```
etobella: remove the `" "`. Probably due to black

etobella (approved):

document_page_approval/hooks.py
flachica: Rebase in progress

jarroyomorales (approved):
LGTM!

Abranes (approved):
LGTM

PR #249, opened 2020-04-11
[ADD] website_document_page

max3903:
Provides a public wiki to host documentation

Depends on:
- [ ] #250


max3903:
Ping @ygol @kittiu

ygol (commented):
- The wiki page mixes both wiki categories and pages. Is this on purpose?

- It seems the category's content is discarded when saving.

- when clicking on a "subcategory" category with a defined parent; you are re-directed to the backend form in place of the frontend.

- There is an access error for guest user

max3903:
@ygol This is a prototype. I am not sure yet what to do and how.

A first version should at least include the following:
* allow unauthenticated users to view pages
* allow authenticated users to edit pages
* provide a configurable menu
* allow to search
* allow to tag and search/filter using tags

PR #250, opened 2020-04-12
[ADD] document_tag

max3903:
Depends on:
- [x] #251


ygol (commented):
wouldn't it make sense to have the tag directly added the document_page? do we need a separate module?

kittiu:
I agree with @ygol to have this part of page.

max3903:
@OCA/document-maintainers

ivantodorovich:
I agree it makes more sense to include this in `document_page` module, than having a separate module.
In any case thanks @max3903 for this :)

etobella:
Which is the sense of this module if there is already `document_page_tag` on 12.0 that has a similar behavior?

https://github.com/OCA/knowledge/tree/12.0/document_page_tag

The only missing thing is company and parent-child behavior, this can be added easily on the original module if it is needed

ivantodorovich:
> Which is the sense of this module if there is already `document_page_tag` on 12.0 that has a similar behavior?
> 
> https://github.com/OCA/knowledge/tree/12.0/document_page_tag
> 
> The only missing thing is company and parent-child behavior, this can be added easily on the original module if it is needed

Indeed, but in any case I'm in favour of merging that feature into `document_page`. That's easier to maintain than a bunch of small modules, specially for non invasive features like this one

etobella:
Ok then! @max3903 Can you check it?

max3903:
@etobella Sorry guys, I don't have time for this.

PR #257, opened 2020-05-26
[FIX][13.0] retrive 12.0 behaviours in category creation without content

Tonow-c2c:
The document.page content for category is required as in 12.0 but in IHM in 12.0 we can create a category without fill this field.

It's only one fix to retrive the same behaviour but investigation in the field definition and computation will be great

* to avoid  to add content (not save in the category creation)
Peek 2020-05-25 16-03

* and create directly the category
Peek 2020-05-26 11-27

### DIFF
--- a/document_page/views/document_page_category.xml
+++ b/document_page/views/document_page_category.xml
@@ -38,6 +38,7 @@
                                 widget="html"
                                 class="oe_view_only"
                                 options='{"safe": True}'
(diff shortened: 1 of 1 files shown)


guewen (approved):

marcelsavegnago (approved):

etobella (approved):
LGTM

PR #267, opened 2020-11-04
[FIX] Don't set `content_parsed` an HTML field or it will be sanitized

### LINKED ISSUE #266: [12.0] document_page_reference: iframe youtube video missing from rendering
@etobella, I didn't take a look on this module code but when you install it, pages that contains links to Youtube videos are not rendered correctly since the iframe tag is no more added to the readonly view.

ypapouin:
Fix #266 
Do not set this field as HTML or it will be sanitized in `convert_to_cache` in `odoo/odoo/fields.py` and iframe tags will be removed

### DIFF
--- a/document_page_reference/models/document_page.py
+++ b/document_page_reference/models/document_page.py
@@ -49,7 +49,10 @@ class DocumentPage(models.Model):
     reference = fields.Char(
         help="Used to find the document, it can contain letters, numbers and _"
     )
-    content_parsed = fields.Html(compute='_compute_content_parsed')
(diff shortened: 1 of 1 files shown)


jarroyomorales:
Hi! I just tested it on runbot and links are not working

etobella (changes requested):
If the problem is the sanitization, you can do:

`content_parsed = fields.Html(compute='_compute_content_parsed', sanitize=False)`

etobella (approved):
After checking how it is managed on parent modules, I think your approach is the right one. Sorry for the misunderstanding :wink:

PR #269, opened 2020-11-05
[IMP] Apply template

yajo:
This repo needs https://github.com/OCA/oca-addons-repo-template/pull/39 to migrate an addon.

@Tecnativa TT19733 @sbidoul @joao-p-marques @victoralmau


sbidoul (approved):

joao-p-marques (approved):

yajo:
I think we should expect the same problem on all repos that don't get the same process... do we have a way to automate it?

sbidoul:
I still need to push the pylint-odoo version patch to 13.0 branches. I do this using [oca-patch-branch](https://github.com/OCA/maintainer-tools/pull/481/files).

I hesitate to apply the copier template to 13.0 branches because
- there are local overrides here and there
- there is the README.md issue (or maybe we can override it?)
- too much trouble on existing PRs?

joao-p-marques:
The copier template may introduce some overrides now, but maybe in the long run it is the best option, and will make it a lot easier to propagate this types of fixes and to keep other "non-essential" configurations up-to-date (like https://github.com/OCA/knowledge/pull/269/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R51), making it easier to work on all repos.

yajo:
> The copier template may introduce some overrides now, but maybe in the long run it is the best option, and will make it a lot easier to propagate this types of fixes and to keep other "non-essential" configurations up-to-date

I agree. However, I know I don't currently have that maintenance burden on my shoulders, so I have to give some extra credits to @sbidoul opinion. Some points on what you mention:

> * there are local overrides here and there

Some specific examples might help. Maybe it can be applied massively for the majority and leave corner cases to be handled manually.

> * there is the README.md issue (or maybe we can override it?)

It wouldn't be the most beautiful solution, but given v14 branches are already fixed, you might get the proper description from there somehow?

> * too much trouble on existing PRs?

I think the biggest problem would be `.pre-commit-config.yaml`, but given the template freezes the 13.0 version, I don't think it would be lots of trouble.

PR #270, opened 2020-11-05
[WIP] 14.0 mig attachment_preview

fshah-initos:
[WIP] 14.0 mig attachment preview


fshah-initos:
Hello all,

For the attachment preview module, I face an issue getting the chatter object in FormRenderer class so we can't set the attachment in the AttachmentPreview widget.
I have pushed the latest code.
can someone help me with this issue.?

attachment_preview/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2014 Therp BV (<http://therp.nl>)
```
NL66278, line 1: Please remove copyright lines from init files and change license lines (in all files) to https:

attachment_preview/models/ir_attachment.py
```diff
@@ -0,0 +1,83 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class IrAttachment(models.Model):
+    _inherit = "ir.attachment"
+
+    @api.model
+    def get_binary_extension(self, model, ids, binary_field, filename_field=None):
+        # import pdb
```
NL66278, line 19: Remove pdb lines

attachment_preview/static/lib/ViewerJS/compatibility.js
```diff
@@ -0,0 +1,577 @@
+/* -*- Mode: Java; tab-width: 2; indent-tabs-mode: nil; c-basic-offset: 2 -*- */
```
NL66278, line 1: This whole file loads a lot of compatibility fixes for versions of browsers and android that are not supported by recent versions of Odoo anyway, and certainly not by 14.0. So I think we should be able to get rid of this whole file.

attachment_preview/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "license": "AGPL-3",
+    "summary": "Preview attachments supported by Viewer.js",
+    "category": "Knowledge Management",
+    "depends": ["web", "mail"],
+    "data": [
+        "templates/assets.xml",
+    ],
+    "qweb": [
+        "static/src/xml/attachment_preview.xml",
+    ],
```
NL66278, line 18: You must add a key for external dependencies and add "magic" to it.

attachment_preview/models/ir_attachment.py
```diff
@@ -0,0 +1,83 @@
...
+                        this[filename_field],
+                    )
+        # Second pass for all attachments which have to be loaded fully
+        #  to get the extension from the content
+        ids_to_browse = [_id for _id in ids_to_browse if _id not in result]
+        for this in self.env[model].with_context(bin_size=True).browse(ids_to_browse):
+            if not this[binary_field]:
+                result[this.id] = False
+                continue
+            try:
```
NL66278, line 52: Move the import of magic to the top of the file, after setting _logger, like so:
```
try:
    import magic
except ImportError:
    _logger.info(_("Library magic is not available, will guess mime types."))
    magic = None
```
Also add import of `_` (translations) from odoo to line importing api and models.
Add translation to all log messages.

In the get_binary_extention method test if magic == None, and in that case use the mimetypes.guess_type() method, instead of doing this in an exception handler. Better by the way to break up this large method in smaller pieces.

NL66278:
And indeed nothing is really shown, to open this. It would be nice if it could be functionally equivalent to what happens in enterprise, that you can view an attachment on the right side of the form/document it is attached to.

NL66278 (changes requested):
Thanks for trying to migrate this, but the web client has changed considerably in 14.0, therefore this will not be working as is.

OSevangelist:
Hi @NL66278 do you have any suggestion for @fshah-initos how or in which direction to change that in order to make it work as you said "in a similar way as enterprise" or do you think this is senseless after all. I honestly don't know why and in relation to what concrete project @fshah-initos initially tried to migrate that but i think function that is merely similar to Odoo v14++ EE showing an attachment right to the data is is linked to would still be desirable but doesn't exist. Am i right ?

espo-tony:
Hi, I believe this module is no longer required. The exact UI and UX are no longer consistent with Odoo layout as it is in v14, and it's functionally equivalent enough to either what base Enterprise already does (for those who use Enterprise) or to what is introduced by [web_responsive](https://github.com/OCA/web/tree/14.0/web_responsive) (for those who don't use Enterprise).

Please, see below an example of the attachment preview functionality in Odoo v14CE with web_responsive installed.
 
web_responsive

vancouver29:
For anyone interested in the `attachment_preview`  module in V14. [PR360](https://github.com/OCA/knowledge/pull/360)

PR #277, opened 2020-12-30
[MIG] 14.0 knowledge and document_page

PierrickBrun:

PR #264, opened 2020-10-18
[14.0][MIG] attachment_category

rousseldenis:

PR #278, opened 2020-12-31
[ADD] 14.0 add knowledge_attachment_category

PierrickBrun:
based on #277 and #264


sebastienbeau:
Hi, I think it's a good time to refactor a little this modules.
I think we should have two thing:
- tags (m2m)
- category (m2o)

On category we should be able to configure access right (restricting to specific user)

Tags are just here to help user for searching

We should be able to filter attachment (and potentially autotag, autocategories) add on a object (product, project....)

PR #289, opened 2021-03-07
[13.0][FIX] document_url: Fix module dependencies.

sergio-teruel:
cc @Tecnativa
ping @CarlosRoca13 @chienandalu

### DIFF
--- a/document_url/__manifest__.py
+++ b/document_url/__manifest__.py
@@ -7,6 +7,7 @@
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/knowledge",
     "license": "AGPL-3",
+    "depends": ["mail"],
     "data": ["view/document_url_view.xml"],
     "qweb": ["static/src/xml/url.xml"],
     "installable": True,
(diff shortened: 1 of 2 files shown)


document_url/__manifest__.py
```diff
@@ -7,6 +7,7 @@
     "author": "Tecnativa, Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/knowledge",
     "license": "AGPL-3",
+    "depends": ["web"],
```
pedrobaeza: Shouldn't it be mail?

sergio-teruel: I do not see that it is necessary... but I am not sure..

pedrobaeza: Seeing this:

https://github.com/OCA/knowledge/blob/f79f13724ff1aed24b83f23b2a79bf42ef22ae16/document_url/static/src/js/url.js#L11

I think it should be.

sergio-teruel:
@pedrobaeza changes done!! :smile:

pedrobaeza (approved):

chienandalu (approved):

PR #292, opened 2021-04-09
[12.0][FIX] document_page_approval: Delete diff in email_template to prevent errors in some cases

victoralmau:
Delete diff in email_template to prevent errors in some cases.

If we added image in base64 inside content:
``
when Odoo try to render message and send try to create `ir.attachment` with `type=content` (it's not possible).

Please @joao-p-marques  and @chienandalu can you review it?

@Tecnativa TT29089

### DIFF
--- a/document_page_approval/data/email_template.xml
+++ b/document_page_approval/data/email_template.xml
@@ -26,11 +26,6 @@
 <p>${object.summary}</p>
 % endif
 
-<h3>Diff</h3>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please be specific about which "some cases". If not, we/others won't remember the reason for doing it. You can put an explanation and relevant part of stack trace.

chienandalu (approved):

joao-p-marques (approved):

pedrobaeza:
Please put the info on the commit

PR #293, opened 2021-04-15
[13.0][FIX] document_page_approval: Delete diff in email_template to prevent errors in some cases

victoralmau:
FWP from 12.0: https://github.com/OCA/knowledge/pull/292

Delete diff in email_template to prevent errors in some cases.

If we added image in base64 inside content:
``
when Odoo try to render message and send try to create `ir.attachment` with `type=content` (it's not possible).

Please @joao-p-marques  and @chienandalu can you review it?

@Tecnativa TT29089

### DIFF
--- a/document_page_approval/data/email_template.xml
+++ b/document_page_approval/data/email_template.xml
@@ -29,11 +29,6 @@
 <p>${object.summary}</p>
 % endif
 
-<h3>Diff</h3>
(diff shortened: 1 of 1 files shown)


joao-p-marques (approved):

victoralmau:
ping @chienandalu

chienandalu (approved):

PR #300, opened 2021-07-24
Updated: mig_document_page_project migrate v14

thomaspot:
Migrated module:   document_page_project

Module was used for v13; Installing, Testing without further actions. Using in porduction env.


gaikaz:
@thomaspot Thank you for contributing.
Before we review, there are a couple of things you didn't do by OCA migration guidelines.
Could you please read this: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-14.0
And then re-commit your work?

Issues in mind:
- [ ] You didn't `git-patch` correctly so all commits of prior versions are missing
- [ ] The version should be changed to `14.0.1.0.0` ("reset" the [semver](https://semver.org) part to "1.0.0")
- [ ] No need to touch the main `README.md` of the repo as it now even conflicts. I suppose maintainers should ask for that change or do it later.

PR #304, opened 2021-09-01
[WIP] [14.0] [MIG] document_url

DavidBForgeFlow:
Standard migration to v14.

PR #306, opened 2021-10-13
[14.0][MIG] document_url: Migration to 14.0

victoralmau:
Supersede from: https://github.com/OCA/knowledge/pull/304

Migration to 14.0.

Extra changes:
- Define the add attachment and add URL buttons on the same line.
- Allow downloading of url type attachments by clicking on the icon or the name.
- Define the name of the url type attachments as links (allows to open it in a new tab).
- Hide the download button for url attachments.
- Force to set `mimetype` to "**application/link**" for url attachments.
- Displays a link icon for url attachments.

Please @Tardo and @pedrobaeza can you review it?

@Tecnativa TT30263


pedrobaeza (changes requested):
Functional review:

- The "Add URL" button link appears in a new line, taking more space, instead of appearing in the same line as previous versions.
- Although it's not explicitly something of this module, the way URL attachments now behave seems odd: they appear as a regular attachment, but with a generic icon (non clickable), and the download button. Only clicking on such button leads to open the URL in the same tab, but it's not possible to do ctrl+click (or third-button click) to open in new tab. I think this module should add some changes to this behavior to have a special icon like in previous versions, and be totally clickable, opening the link in a new tab.

pedrobaeza:
This is better, thanks, but I'm still unable to open it in a new tab, and it's opened by default in the same tab. The download icon can be also hidden in this type of attachment. A last thing: about the icon, isn't a more suitable icon for the link?

victoralmau:
Some changes have been made.

There are other pending changes that would be quite difficult to make (it would require overwriting quite a bit of core code):
- Change the icon of url type attachments (in v13 it was the same).
- Change the icon and define it as a link (now an `on-click` is defined that does not allow it to be opened in a new tab).

pedrobaeza:
Then what are the changes you have made? Are you going to handle the rest? At least we need to do ctrl+click for opening in new tab.

victoralmau:
> Then what are the changes you have made? Are you going to handle the rest? At least we need to do ctrl+click for opening in new tab.

Sorry, I have not mentioned the changes (if they are in the commit).
The changes that are already made are the following:
- Hide the download button in url type attachments.
- Define the name of the attachment as a link, this allows to open it in a new tab with Ctrl + Click.

pedrobaeza:
This is better, thanks, although the other features are very desired. Isn't possible to add to the link (`<a>` tag?) the attribute `target="new"` for having the new tab thing? Another thing is that the target should be the URL, not the controller path that is later resolved to the link.

victoralmau:
Some more clarifications regarding the previous comment::
- Yes, it is possible to add `target="new"` to the link.
- It is not possible to add `attachment.url` directly because that field was not get from https://github.com/odoo/odoo/blob/14.0/addons/mail/static/src/models/thread/thread.js#L680
- The icon cannot be changed (it is not really an icon, it is a div that has as its `backround` the `mimetype` https://github.com/odoo/odoo/blob/14.0/addons/mail/static/src/components/attachment /attachment.xml#L17-L25)

I think one possibility would be to suggest some changes to Odoo to help with this, what do you think?

pedrobaeza:
Odoo is not going to change anything about this, as URLs are not allowed anywhere in the UI. Aren't you able to intercept the mimetype compute method for putting a special one, and provide the mimetype icon here in this module instead on the main one?

victoralmau:
I have added all the extra changes to the PR description to clarify all of them.

pedrobaeza:
Great, we are advancing a lot, but now you have hidden too much, as for PDFs, the download button nor the extension is shown (after refreshing one time the tab):

On this PR:
Selección_051

On runbot:

Selección_050

pedrobaeza:
Now the download button has appeared again for links.

victoralmau:
> Now the download button has appeared again for links.

Sorry for the absurd confusion, it's already changed.
adjuntos

pedrobaeza (approved):
Now it's working as desired.

Tardo (approved):
LGTM :+1: Only code review.

The mimetype for attachments of type url seems not to be relevant. The field is calculated when you write/create and on the other hand the mimetype 'application/link' doesn't seem to be valid (https://www.iana.org/assignments/media-types/media-types.xhtml). So if you really want to force a mimetype... it might be more convenient to use application/octet-stream

pedrobaeza:
Let's move on for now:

/ocabot merge nobump

I think the mimetype helps for displaying such icon.

CasVissers-360ERP:
@pedrobaeza Merge seems stalled?

PR #331, opened 2022-03-15
[14.0][ADD] attachment_zipped_download: New addon.

victoralmau:
This module allows downloading multiple attachments as a zip file.

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT33403


attachment_zipped_download/models/ir_attachment.py
```diff
@@ -0,0 +1,36 @@
...
+
+from odoo import _, models
+from odoo.exceptions import UserError
+
+
+class IrAttachment(models.Model):
+    _inherit = "ir.attachment"
+
+    def action_attachments_download(self):
+        items = self.filtered(lambda x: x.type == "binary")
```
pedrobaeza, line 15: Why filter on binary?

attachment_zipped_download/models/ir_attachment.py
```diff
@@ -0,0 +1,36 @@
...
+            "url": "/web/attachment/download_zip?ids=%s" % (ids),
+            "target": "self",
+        }
+
+    def _create_temp_zip(self):
+        zip_buffer = BytesIO()
+        with zipfile.ZipFile(zip_buffer, "a", zipfile.ZIP_DEFLATED, False) as zip_file:
+            for attachment in self:
+                zip_file.write(
+                    attachment._full_path(attachment.store_fname), attachment.name
```
pedrobaeza, line 32: Why full_path?

attachment_zipped_download/models/ir_attachment.py
```diff
@@ -0,0 +1,36 @@
...
+        ids = ",".join(map(str, items.ids))
+        return {
+            "type": "ir.actions.act_url",
+            "url": "/web/attachment/download_zip?ids=%s" % (ids),
+            "target": "self",
+        }
+
+    def _create_temp_zip(self):
+        zip_buffer = BytesIO()
+        with zipfile.ZipFile(zip_buffer, "a", zipfile.ZIP_DEFLATED, False) as zip_file:
```
pedrobaeza, line 29: This means no compression, isn't it? Put on ROADMAP to make this configurable.

pedrobaeza (commented):
Missing usage instructions (on the system attachments menu).

attachment_zipped_download/models/ir_attachment.py
victoralmau: Not exactly, the explanation from the https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile documentation is as follows:

_compression is the ZIP compression method to use when writing the archive, and should be ZIP_STORED, ZIP_DEFLATED, ZIP_BZIP2 or ZIP_LZMA; unrecognized values will cause NotImplementedError to be raised. If ZIP_DEFLATED, ZIP_BZIP2 or ZIP_LZMA is specified but the corresponding module (zlib, bz2 or lzma) is not available, RuntimeError is raised. The default is ZIP_STORED._

attachment_zipped_download/models/ir_attachment.py
victoralmau: The idea is to exclude files of type url.

attachment_zipped_download/models/ir_attachment.py
victoralmau: It is necessary to specify the path so that the file is added correctly because the file is not in the same folder from where the .zip file is created.

pedrobaeza (approved):

chienandalu (approved):
Test and code review :+1:

PR #345, opened 2022-05-09
[14.0][FIX]: change the priority views form.

angelgarciadelachica:
When we plan an activity in a page and from the kanban view of planned activities we click on an activity it loads the document.page.category.form view instead of document.page.form view. 
The issue is solved by changing the priorities of both views.

PR #348, opened 2022-05-18
[15.0][FIX] document_page: change the priority views form.

angelgarciadelachica:
When we plan an activity in a page and from the kanban view of planned activities we click on an activity it loads the document.page.category.form view instead of document.page.form view.
The issue is solved by changing the priorities of both views.
Same fix as in v14. #345

### DIFF
--- a/document_page/views/document_page.xml
+++ b/document_page/views/document_page.xml
@@ -33,6 +33,7 @@
     <record id="view_wiki_form" model="ir.ui.view">
         <field name="name">document.page.form</field>
         <field name="model">document.page</field>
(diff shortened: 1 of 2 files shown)


angelgarciadelachica:
@HaraldPanten @ValentinVinagre

ValentinVinagre:
@etobella Es lo mismo que el de la v.14 https://github.com/OCA/knowledge/pull/345

ValentinVinagre (approved):
LGTM 👍🏻

HaraldPanten (approved):

PR #357, opened 2022-07-26
[15.0][MIG] document_page_approval: Migration to 15.0

victoralmau:
Migration to 15.0

Please @pedrobaeza and @ernestotejeda can you review it?

@Tecnativa TT36950


document_page_approval/models/document_page.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class DocumentPage(models.Model):
+    """Useful to know the state of a document."""
+
+    _inherit = "document.page"
+
+    history_ids = fields.One2many(
+        order="approved_date DESC", domain=[("state", "=", "approved")]
+    )
```
FernandoRomera: _order="approved_date DESC"

history_ids = fields.One2many(
         domain=[("state", "=", "approved")]
    )

victoralmau: `order` or `_order` attribute does not work on a field.
I have added this ordering to the tree view with `default_order` attribute.

FernandoRomera: I think there was a mistake. To correct: remove the field order attribute and add the _order attribute at the table level.

pedrobaeza: But that's not the intention. The goal is to order the children model, so the other technique is the required one.

document_page_approval/models/document_page.py
```diff
@@ -48,6 +46,7 @@ class DocumentPage(models.Model):
         "Approval required",
         help="If true, changes of this page require approval",
         compute="_compute_is_approval_required",
+        recursive=True,
```
pedrobaeza, line 49: Why do you need this?

document_page_approval/models/document_page.py
```diff
@@ -57,17 +56,20 @@ class DocumentPage(models.Model):
         string="Approver groups",
         help="Groups that can approve changes to this document",
         compute="_compute_approver_group_ids",
+        recursive=True,
```
pedrobaeza, line 59: And this?

document_page_approval/models/document_page.py
```diff
@@ -57,17 +56,20 @@ class DocumentPage(models.Model):
...
     )
 
     has_changes_pending_approval = fields.Boolean(
         compute="_compute_has_changes_pending_approval",
-        string="Has changes pending approval",
```
pedrobaeza: Isn't this changing the casing of the string to `Has Changes Pending Approval`?

document_page_approval/models/document_page_history.py
```diff
@@ -11,6 +11,7 @@ class DocumentPageHistory(models.Model):
 
     _name = "document.page.history"
     _inherit = ["document.page.history", "mail.thread"]
+    _order = "id DESC"
```
pedrobaeza: Shouldn't this be ordered by `approved_date`?

document_page_approval/models/document_page_history.py
```diff
@@ -126,8 +127,8 @@ def action_cancel(self):
         for rec in self:
             rec.message_post(
                 subtype_xmlid="mail.mt_comment",
-                body=_("Change request <b>%s</b> has been cancelled by %s.")
-                % (rec.display_name, self.env.user.name),
+                body=_("Change request <b>%(name)s</b> has been cancelled by %(user)s.")
```
pedrobaeza, line 130: Change translations for this, at least on Spanish

document_page_approval/models/document_page.py
victoralmau: Without this change, when executing the tests, the following log is shown:

`WARNING devel py.warnings: /opt/odoo/custom/src/odoo/odoo/fields.py:722: UserWarning: Field document.page.approver_group_ids should be declared with recursive=True`

document_page_approval/models/document_page.py
victoralmau: Without this change, when executing pre-commit the following log is shown:

`The attribute string is redundant. String parameter equal to name of variable`

pedrobaeza: Yeah, but that's not my question? If you remove it, what is the label? If it's with capital letters, then the pylint check is not correct. If it's the same, then my bad and everything is correct.

document_page_approval/models/document_page.py
pedrobaeza: Uhm, OK, but I don't know why

document_page_approval/models/document_page.py
victoralmau: Ok sorry, removing the string tag odoo auto-defines it with "_Has Changes Pending Approval_" (wrong). The pylint validation seems to be wrong.

pedrobaeza: Please put the issue in pylint-odoo project.

pedrobaeza (approved):

victoralmau:
Ping @ernestotejeda

victoralmau:
Ping @chienandalu

chienandalu (approved):
Tested :+1:

PR #358, opened 2022-08-01
13.0 mig attachment preview

vancouver29:
migration of attachment_preview from v12 to v13


vancouver29:
attachment_preview Migration to 13.0

@pedrobaeza and @ernestotejeda could you please review it? Travis failed because of the `black` hook at pre-commit!

@rousseldenis could you please check the pre-commit error?

pedrobaeza:
Sorry, not interested in this module anymore.

rousseldenis:
> attachment_preview Migration to 13.0
> 
> @pedrobaeza and @ernestotejeda could you please review it? Travis failed because of the `black` hook at pre-commit!
> 
> @rousseldenis could you please check the pre-commit error?

You should install pre-commit locally before committing. This will reformat correctly automatically. Or run `pre-commit run -a` to do it manually

rousseldenis:
@vancouver29 Thanks for this. Could you follow migration guide, especially full process: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0

Reformatting changes should appear in the preceding commit (black, isort, ...) as only migration changes should appear in the ... migration one :smiley:

vancouver29:
> @vancouver29 Thanks for this. Could you follow the migration guide, especially the full process: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0
> 
> Reformatting changes should appear in the preceding commit (black, isort, ...) as only migration changes should appear in the ... migration one smiley
>
@rousseldenis, thanks for the feedback :) I will remember it next time. Could anyone please review my PR? I would love to finish this soon so that I can continue the migration of this module to the following versions, 14 and 15

.travis.yml
```diff
@@ -21,7 +21,7 @@ jobs:
   include:
     - stage: linting
       name: "pre-commit"
-      install: pip install pre-commit
+      install: pip install pre-commit python-magic
```
rousseldenis: @vancouver29 I'm wondering why adding this ?

vancouver29: @rousseldenis, after the Travis test is executed I got the error that `magic` is not available. Hence, that is why I added `python-magic` there

rousseldenis: Where is it required ?

vancouver29: In the `ir_attachment.py` file on lines 51, 54, and 58 :)

rousseldenis: Ok, see, looked into files headers.

You should add an external_dependencies to that library in module manifest instead.

vancouver29: @rousseldenis I tried to add either `magic` or `python-magic` to `external_dependencies` in the manifest, but the test seems to be failed and it does not recognize this dependency. Hence, i think the only way to make it work here is to add `python-magic` into the `.travis.yml` file. Or do you have any idea?

rousseldenis: > Or do you have any idea?

Yes. The good one is to switch to Github Actions and then, setuptools-odoo will detect the entry in external_dependencies.

I'll do the PR for that.

rousseldenis: @vancouver29 #359

vancouver29: @rousseldenis Great :)! I'm looking forward that your PR being merged, then I can finish this PR and start the migration of this module for V14 & V15

Issue #366, opened 2022-10-03
Migration to version 16.0


sbidoul:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

# Modules to migrate

- [ ] attachment_preview - By @houzefa-abba - #434
- [x] attachment_zipped_download - By @FernandoRomera - #391
- [x] document_page - By @etobella - #382
- [x] document_page_access_group - By @Tisho99 - #470
- [x] document_page_approval - By @FernandoRomera - #388
- [x] document_page_group - By @FernandoRomera - #383
- [x] document_page_partner - By @ntsirintanis - #465
- [x] document_page_project - By @ntsirintanis - #464
- [x] document_page_reference - By @kaynnan - #422
- [x] document_page_tag - By @FernandoRomera - #384
- [x] document_url - By @FernandoRomera - #385
- [x] knowledge > document_knowledge - By @marielejeune - #368

Missing module? Check https://github.com/OCA/maintainer-tools/wiki/%5BFAQ%5D-Missing-modules-in-migration-issue-list

HarhuTechnologies:
Hi @sbidoul : I want raise a PR against these modules. I have already migrated. Can you please help me how can I create a PR against this issue? @pedrobaeza

pedrobaeza:
Do the pull request, and someone will link through the bot command the PR in this issue. Please follow the migration guide for proper migration mechanism.

pedrobaeza:
Another thing: `knowledge` must be renamed to `knowledge_oca` or similar for not collapsing with Odoo new enterprise module.

HarhuTechnologies:
alright @pedrobaeza : Thanks for letting me know. I will create PR.

pegonzalezspesol:
working on document_page

vancouver29:
> # Todo
> https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0
> 
> # Modules to migrate
> * [ ]  attachment_zipped_download - By @FernandoRomera - [[16.0][MIG] attachment_zipped_download: Migration to 16.0 #391](https://github.com/OCA/knowledge/pull/391)
> * [x]  document_page - By @etobella - [[16.0][MIG] document_page: Migration to 16.0 #382](https://github.com/OCA/knowledge/pull/382)
[...]

PR #367, opened 2022-10-19
[16.0] [MIG] knowledge

marielejeune:
Migrated to version 16.0

TODO: change module name, since knowledge is now a distinct Odoo module.


marielejeune:
https://github.com/OCA/knowledge/issues/366

pedrobaeza:
For having the whole history with the new name for later migrations (in v17, v18...), please execute this command:

```bash
git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD
```

marielejeune:
@pedrobaeza for changing module name, do you see something more to do than running https://github.com/OCA/openupgradelib/blob/94762d1a5a4172b680a7e019fe4d7db44b2bf7af/openupgradelib/openupgrade.py#L1529 in a pre-migration script?

pedrobaeza:
Hi, @marielejeune 

That's only needed when you are migrating from 15.0 to 16.0, but not for the module migration itself. On OpenUpgrade, we add all the renamings/mergings in this file:

https://github.com/OCA/OpenUpgrade/blob/15.0/openupgrade_scripts/apriori.py

and the method you mentioned is called from `base` module migration script.

pedrobaeza:
Good name selected. Please remember to rename whole commit history.

sbidoul:
> Please remember to rename whole commit history.

Ow, that is surprising. I understand the reason but... one more thing to add to my list of downsides of our current migration approach ;)

@sebalix as you work on oca-port, what do you think of this idea of renaming a module in past commits when migrating? Have you already thought about module renaming in oca-port?

pedrobaeza:
The problem will be similar when reusing the same branches, as git wont' find references before the renaming, so I think this is even better, as you have the occasion to accommodate the history for this case.

marielejeune:
Hi @pedrobaeza, I'm not sure to understand what this command should do?
`git filter-branch --tree-filter 'if [ -d old_module_name ]; then mv old_module_name new_module_name; fi' HEAD`

I launched it locally but apart new commits sha, I don't see any difference.

pedrobaeza:
You need to change `old_module_name` and `new_module_name` by the corresponding ones: `knowledge` and `document_knowledge`.

marielejeune:
Yes, that was done

pedrobaeza:
Well, if you inspect a commit before the last one, check if now the module directory has the new name.

pedrobaeza:
You can forced push it here also to be checked.

pedrobaeza:
Oh, I think you need to do a rebase over origin/16.0 before...

sebalix:
@sbidoul I thought about a  `--rename-to` option but nothing done right now. I didn't know we should rename the module also in the past commits.
I created https://github.com/OCA/oca-port/issues/11

sbidoul:
> I didn't know we should rename the module

Well I'm not sure we should either, TBH.

PR #372, opened 2022-11-16
[14.0][IMP] document_page_reference: auto-fill reference when not supplied.

LoisRForgeFlow:
also, make reference optional in tree view and not the first field.

@ForgeFlow


etobella (approved):

LoisRForgeFlow:
A small test has been added.

JordiBForgeFlow (approved):

PR #386, opened 2022-12-08
[FIX] document_knowledge: Error typo res_config

FernandoRomera:


setup/_metapackage/setup.py
```diff
@@ -9,6 +9,7 @@
     version=version,
     install_requires=[
         'odoo-addon-document_knowledge>=16.0dev,<16.1dev',
+        'odoo-addon-document_page>=16.0dev,<16.1dev',
```
pedrobaeza: This shouldn't go here

AntoniRomera (approved):

MRomeera (approved):
ok

max3903 (approved):

setup/_metapackage/VERSION.txt
```diff
@@ -1 +1 @@
-16.0.20221111.0
\ No newline at end of file
+16.0.20221207.0
```
pedrobaeza: And why this change?

FernandoRomera: @pedrobaeza 
I don't know, I guess some process must create the setup directory.
How do you think it would be better to act: recover the previous value, or launch some command to correct it.

pedrobaeza: Rebase and remove this change

pedrobaeza (changes requested):
It includes a merge commit in the branch.

setup/_metapackage/VERSION.txt
pedrobaeza: This is still happening.

PR #389, opened 2022-12-14
[16.0][FIX] document_knowledge: Error in data-key res_config

FernandoRomera:


document_url/i18n/am.po
```diff
@@ -24,14 +24,26 @@ msgid "Add"
...
-#. openerp-web
+#. odoo-javascript
 #: code:addons/document_url/static/src/xml/url.xml:0
 #: model:ir.actions.act_window,name:document_url.action_ir_attachment_add_url
 #: model_terms:ir.ui.view,arch_db:document_url.view_ir_attachment_add_url_form
 #, python-format
 msgid "Add URL"
 msgstr ""
 
+#. module: document_url
```
pedrobaeza, line 35: This change doesn't belong here.

FernandoRomera: This seems to have been done by the weblate user (automatically), we have not modified the file. It is the second time that I close a PR (previously for a problem with the version.txt of the setup directory). If you consider it appropriate, I will close it and do it again from scratch.

pedrobaeza: The change is explicitly done on your commit, not by anything more, so you are mixing something at git side on your local. If you prefer, I can do the change for closing this topic.

FernandoRomera: @pedrobaeza 
I would greatly appreciate it.

pedrobaeza:
Superseded by #390

PR #393, opened 2022-12-20
[15.0][IMP] document_page_reference: auto-fill reference when not supplied.

LoisRForgeFlow:
also, make reference optional in tree view and not the first field.

Forward port of #372


MiquelRForgeFlow (approved):
You should mark in the title that this PR is a forward port.

etobella (approved):
/ocabot merge patch

PR #405, opened 2023-03-08
[16.0][FIX] document_url: Change mimetype only if type=url

victoralmau:
FWP from 15.0: https://github.com/OCA/knowledge/pull/404

Other modules like web_pwa_oca uses url field for storing an indexed path for later being searched in the controller, so let's restrict the mimetype change if both url and type=url is set.

Please @pedrobaeza and @sergio-teruel can you review it?

@Tecnativa


pedrobaeza (approved):
Thanks, fast-tracking this patch for avoiding the problem:

/ocabot merge patch

PR #411, opened 2023-04-28
[14.0][FIX] attachment_zipped_download: zip allowed document only

petrus-v:
The previous code allowed any authenticated to retreive any attachment present on odoo filesystem. So a WMS user could technically spoken download a zip with accounting documents.

This implementation is calling read() method to retreive file to archive which have two benefits:

* call the check(read) implements on ir.attachemnt
* (not test) should works with attachment saved somewhere else than the local filesystem (s3 storage, pgsql large object storage...)

### DIFF
--- a/attachment_zipped_download/models/ir_attachment.py
+++ b/attachment_zipped_download/models/ir_attachment.py
@@ -28,9 +28,16 @@ class IrAttachment(models.Model):
         zip_buffer = BytesIO()
         with zipfile.ZipFile(zip_buffer, "a", zipfile.ZIP_DEFLATED, False) as zip_file:
             for attachment in self:
-                zip_file.write(
-                    attachment._full_path(attachment.store_fname), attachment.name
+                attachment.check("read")
+                zip_file.writestr(
+                    attachment._compute_zip_file_name(),
+                    attachment.raw,
                 )
             zip_buffer.seek(0)
             zip_file.close()
         return zip_buffer
+
+    def _compute_zip_file_name(self):
+        """Give a chance of easily changing the name of the file inside the ZIP."""
+        self.ensure_one()
+        return self.name
--- a/attachment_zipped_download/tests/test_attachment_zipped_download.py
+++ b/attachment_zipped_download/tests/test_attachment_zipped_download.py
@@ -1,12 +1,30 @@
 # Copyright 2022-2023 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 import base64
+from unittest import TestCase
 
-import odoo.tests
-from odoo.tests import new_test_user
+from odoo.exceptions import AccessError
+from odoo.tests import HttpCase, SavepointCase, new_test_user
 
 
-class TestAttachmentZippedDownload(odoo.tests.HttpCase):
+class TestAttachmentZippedDownloadBase(TestCase):
+    @classmethod
+    def _create_attachment(cls, env, user, name, model=False, res_id=False):
+        return (
+            env["ir.attachment"]
+            .with_user(user)
+            .create(
+                {
+                    "name": name,
+                    "datas": base64.b64encode(b"\xff data"),
+                    "res_model": model,
+                    "res_id": res_id,
+                }
+            )
+        )
+
+
+class TestAttachmentZippedDownload(HttpCase, TestAttachmentZippedDownloadBase):
     def setUp(self):
         super().setUp()
         ctx = {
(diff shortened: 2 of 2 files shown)


petrus-v:
@pedrobaeza @victoralmau I suppose you should care about this PR regarding access privileged to attachments according the context your are using this module !

attachment_zipped_download/tests/test_attachment_zipped_download.py
```diff
@@ -2,11 +2,26 @@
...
 import base64
 
-import odoo.tests
-from odoo.tests import new_test_user
+from odoo.exceptions import AccessError
+from odoo.tests import HttpCase, SavepointCase, new_test_user
 
 
-class TestAttachmentZippedDownload(odoo.tests.HttpCase):
+def create_attachment(env, user, name, model=False, res_id=False):
```
pedrobaeza: Why doing this outside the test class? That way, you can't inherit it.

attachment_zipped_download/models/ir_attachment.py
```diff
@@ -28,9 +28,18 @@ def _create_temp_zip(self):
...
+                zip_file.writestr(
+                    attachment._compute_zip_file_name(),
+                    attachment.read(["raw"])[0]["raw"],
                 )
             zip_buffer.seek(0)
             zip_file.close()
         return zip_buffer
+
+    def _compute_zip_file_name(self):
+        """Give a change to easly change name of file inside zip"""
```
pedrobaeza:
```suggestion
        """Give a chance of easily changing the name of the file inside the ZIP."""
```

attachment_zipped_download/models/ir_attachment.py
```diff
@@ -28,9 +28,18 @@ def _create_temp_zip(self):
         zip_buffer = BytesIO()
         with zipfile.ZipFile(zip_buffer, "a", zipfile.ZIP_DEFLATED, False) as zip_file:
             for attachment in self:
-                zip_file.write(
-                    attachment._full_path(attachment.store_fname), attachment.name
+                # Use attachment read for two purpose (initial implementation wasn't):
+                # * to ensure access write
+                # * do not assume file is stored on filesystem
+                zip_file.writestr(
```
pedrobaeza, line 32: Are you sure `writestr` is compatible with all the contents?

attachment_zipped_download/models/ir_attachment.py
```diff
@@ -28,9 +28,18 @@ def _create_temp_zip(self):
...
+                    attachment._compute_zip_file_name(),
+                    attachment.read(["raw"])[0]["raw"],
```
pedrobaeza: Not sure `read` is better than `attachment.datas`

attachment_zipped_download/tests/test_attachment_zipped_download.py
petrus-v: To make it clear that we can use it from class / instance method and also because I've reused it in next PR #412 but your are right probably proprer to add a common TestCase. I'll refactor that a bit.

attachment_zipped_download/models/ir_attachment.py
petrus-v: [Documentation tells that it can receive str or bytes](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.writestr), I have compressed various types likes  `.jpg`

attachment_zipped_download/tests/test_attachment_zipped_download.py
pedrobaeza: Yes, the way to do it is to put a base test class without `test_*` methods, and in the same module, inherit that base class for the module tests, and you can use the base class in other dependent modules. Example:

https://github.com/OCA/bank-payment/blob/ededf771b02135ad23c888ed4103cf441bc6bc54/account_payment_order/tests/test_payment_order_inbound.py#L14
https://github.com/OCA/bank-payment/blob/ededf771b02135ad23c888ed4103cf441bc6bc54/account_payment_order_grouped_output/tests/test_payment_order_inbound_grouped.py#L12

attachment_zipped_download/models/ir_attachment.py
petrus-v: I'll give a try to use `datas` instead of `raw`. Also I'll re-try to direct access to the record attribute but I believes that the added test will failed and we wont call the check('read') ...

petrus-v: I confirm that test will faill with `attachment.datas` so I can either use:
`attachement.check("read")` before or in one line `attachment.read(["datas"])[0]["datas"]`. Do you have any preferences?  explicity calling attachement.check("read") seems probably easier to read and understand :thinking: 

I'll retry to various type of files to make sure it works with binary and text files...

pedrobaeza: Yes, it's better to call it explicitly for clarity and for reusing cache. I think it's not called because it's already in cache.

petrus-v: done ! And I confirm it's better to use `raw` because `datas` convert content to base64 which is not desired before zipping file !

petrus-v: I've test with various format: .jpeg, .png, .pdf, .md, .cfg, .xlsx and were able unzip and read files afterwards

pedrobaeza:
Can you please squash commits?

petrus-v:
> Can you please squash commits?

done !

pedrobaeza (approved):

victoralmau (approved):

PR #425, opened 2023-08-10
[14.0][FIX] document_page: Stored XSS

SergiCForgeFlow:
Knowledge pages are currently vulnerable to stored XSS attack. This fixes this vulnerabilty changing **content** field from type **Text** to type [Html](https://github.com/odoo/odoo/blob/14.0/odoo/fields.py) so it receives default sanitize.


max3903 (commented):
:+1:

PR #426, opened 2023-08-24
[14.0][FIX] document_page: Stored XSS

SergiCForgeFlow:
Currently Knowledge pages are vulnerable to stored XSS attacks. This is because the 'content' field is type **Text**, so it doesn't receive the default sanitize provided by **Html** fields. 

This PR solves this, changing 'content' field to type **Html**.

### DIFF
--- a/document_page_reference/tests/test_document_reference.py
+++ b/document_page_reference/tests/test_document_reference.py
@@ -30,9 +30,6 @@ class TestDocumentReference(TransactionCase):
         self.page2.write({"reference": False})
         self.assertEqual(self.page1.reference, self.page2.reference)
 
-    def test_check_raw(self):
-        self.assertEqual(self.page2.display_name, self.page1.get_raw_content())
-
     def test_check_reference(self):
         self.assertRegex(self.page1.content_parsed, ".*%s.*" % self.page2.display_name)
 
--- a/document_page/__manifest__.py
+++ b/document_page/__manifest__.py
@@ -4,7 +4,7 @@
 
 {
     "name": "Document Page",
-    "version": "14.0.1.2.3",
+    "version": "14.0.1.2.4",
     "category": "Knowledge Management",
     "development_status": "Production/Stable",
     "author": "OpenERP SA, Odoo Community Association (OCA)",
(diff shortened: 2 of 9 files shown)


document_page_reference/tests/test_document_reference.py
```diff
@@ -30,8 +30,8 @@ def test_no_contrains(self):
         self.page2.write({"reference": False})
         self.assertEqual(self.page1.reference, self.page2.reference)
 
-    def test_check_raw(self):
-        self.assertEqual(self.page2.display_name, self.page1.get_raw_content())
+    # def test_check_raw(self):
+    #     self.assertEqual(self.page2.display_name, self.page1.get_raw_content())
```
max3903:
```suggestion
```

max3903 (approved):

OriolMForgeFlow (approved):
LGTM!

pedrobaeza (changes requested):
All existing contents that are not HTML will show incorrectly after this change. You need to incorporate a migration script that converts them to HTML.

PabloEForgeFlow:
It might be worth adding a test which verifies that the content is indeed sanitized after writing the record.

SergiCForgeFlow:
> All existing contents that are not HTML will show incorrectly after this change. You need to incorporate a migration script that converts them to HTML.

All existing contents currently are stored in a HTML format. Regardless of whether the Field is type Text, the editor takes the input, generates and stores the data in a HTML format.

pedrobaeza (approved):
/ocabot merge nobump

SergiCForgeFlow:
> > All existing contents that are not HTML will show incorrectly after this change. You need to incorporate a migration script that converts them to HTML.
> 
> All existing contents currently are stored in a HTML format. Regardless of whether the Field is type Text, the editor takes the input, generates and stores the data in a HTML format.

Field 'content' on 'document.page' model is compute. So when Field Type is changed to HTML this receives default sanitize. This data is computed from 'content' field on model 'document.page.history', which is stored. If some 'content' data was malicious the malicious script is still stored on the database on model 'document.page.history'. That's why I have added post-migration script on the recent commit. Just by reassigning the content data, now this field also receives default HTML sanitize and the script is completely erased from database.

PR #430, opened 2023-09-08
[16.0][FIX] attachment_zipped_download: create temporary file to stream

### LINKED ISSUE #406: [16.0] attachment_zipped_download: Incomplete file
Failed to open the downloaded zips

## Module
attachment_zipped_download

## Describe the bug
Go to Settings > Technical > Database Structure > Attachments and select some files. Then download files and we can see the error in the downloaded file:
image

## To Reproduce
Tested in runbot 16.0, in 14.0 this run fine.

https://user-images.githubusercontent.com/11718836/224478304-59a1c32c-d3f0-4812-9ada-85075f830c66.mp4

**Expected behavior**

**Additional context**
Tested in Brave 1.48, Chrome 110.0.5481 and Firefox 110.01

len-foss:
Despite accurately getting the size from the BytesIO object, Werkzeug doesn't stream its content.
This updates the code to use Odoo's http.Stream class.

Fixes https://github.com/OCA/knowledge/issues/406

### DIFF
--- a/attachment_zipped_download/controllers/main.py
+++ b/attachment_zipped_download/controllers/main.py
@@ -13,9 +13,11 @@ class AttachmentZippedDownloadController(http.Controller):
             return
         list_ids = map(int, ids.split(","))
         out_file = request.env["ir.attachment"].browse(list_ids)._create_temp_zip()
-        return http.send_file(
-            filepath_or_fp=out_file,
+        stream = http.Stream(
+            type="data",
+            data=out_file.getvalue(),
             mimetype="application/zip",
             as_attachment=True,
-            filename=_("attachments.zip"),
+            download_name=_("attachments.zip"),
         )
+        return stream.get_response()


pedrobaeza (approved):
Tested and working

len-foss:
Is there any reason not to merge this?

petrus-v (commented):
Code looks good but I've the feeling it's a shame here, we have a new class to manage stream and we are not using it...

`_create_temp_zip()` return a `BytesIO` the `getvalue()` return the raw bytes which is then transform again as a bytesIO by the `htpp.Stream` class. In a large zip file this could be a lot of memory.

The title told us that it create a temporary file to stream but I don't think it's the case with this change. I've the feeling it would be probably better to do that, change a bit _create_temp_zip to write a temporary file and use Stream with `type="path"`

As far i understand this mainly avoid to use a deprecated method that I won't be opposed, the improvement can come later !

len-foss:
The history is lost because I force-pushed the branch (a long time ago so it was pruned); the first version used `fo = tempfile.NamedTemporaryFile(); fo.write(out_file.getvalue())` and passed `fo` as the stream argument.
What bothered me is that in any case we have a call to `getvalue`.
If there's a better way we can go for it, I just didn't find one.

HekkiMelody (approved):
Code review

PR #431, opened 2023-09-08
[16.0] pdf content indexing: PyMupdf + Tesseract

len-foss:
It integrates PyMuPDF to perform text extraction.

The OCR with tesseract has also been migrated from version 8; to be able to perform extraction on long documents, PyMuPDF is used to split the content into multiple images, as it has a limit on the image size it can process. I've also added the possibility to explicitly change Tesseract's base language with a context key.

I also adds a new module to perform OCR in individual jobs rather than a cron.


len-foss:
@agent-z28 FYI

PR #432, opened 2023-09-17
[16.0]ADD Partner field

flotho:
IMP: small display improvements

### DIFF
--- a/document_page/models/document_page.py
+++ b/document_page/models/document_page.py
@@ -93,6 +93,12 @@ class DocumentPage(models.Model):
         help="Use it to link resources univocally",
         compute="_compute_backend_url",
     )
+    partner_id = fields.Many2one(
+        comodel_name="res.partner",
+        string="Partner",
+        index=True,
+        ondelete="cascade",
+    )
 
     @api.depends("menu_id", "parent_id.menu_id")
     def _compute_backend_url(self):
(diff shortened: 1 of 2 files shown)


pedrobaeza (changes requested):
This doesn't have too much sense for a general module IMO. Do this in a separate module.

flotho:
Hi @pedrobaeza thanks for your comment.
I'll separate the behaviour in another module.
FMI, what criteria makes you consider that it's not relevent to have this in document page?

pedrobaeza:
This module is for creating document pages, normally for your own company as documentation, management system, etc. On that context, a partner field is not adding value and being misleading.

And another thing: if you add things in an existing module, the label should be `[IMP]`, not `[ADD]`. The second one is for adding a new module. If not, being philosophical, everything will be `[ADD]`, even a fix, as it's adding value to the module, hehe.

flotho:
hum, thanks for the feedback @pedrobaeza 
Regarding the functionnal part, loks like your point of view will lead us to a business requirements insted of a partner field on this module.
My idea was : 

- find easily documentation relevant to a specific customer
- share easily this documentation on the portal

What would be your recommendation ?

regards

pedrobaeza:
For the second point, maybe it's better to use `dms` module. If not, you should add the portal view page visualization with permissions.

flotho:
thanks for sharing your advices.
Regards

PR #440, opened 2023-11-06
[16.0][ADD]document_url_google_drive

DemchukM:
…ach directly from Google Drive


pedrobaeza:
Thanks for the contribution. This should be `[ADD]`, not `[IMP]`, as it's adding a new module, not improving an existing one. Ref: https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#71commit-message

On the other side, please check pre-commit.

OpenCode:
@DemchukM Why there is a deb file? `pandoc-3.1.9-1-amd64.deb`

PR #441, opened 2023-11-06
Updated from template

ivs-cetmix:
Applied latest template


ivs-cetmix:
Hey @pedrobaeza  could you please check and merge it?

attachment_zipped_download/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
```
pedrobaeza: Why these files here?

pedrobaeza: cc @sbidoul is the new system contaminating each module with this file?

ivs-cetmix: > Why these files here?

Sorry no idea) Maybe @sbidoul  could give some hint

.copier-answers.yml
```diff
@@ -19,6 +20,6 @@ repo_description: 'TODO: add repo description.'
 repo_name: knowledge
 repo_slug: knowledge
 repo_website: https://github.com/OCA/knowledge
-travis_apt_packages: []
-travis_apt_sources: []
+use_pyproject_toml: true
```
etobella: Use false here in order to remove that files

pedrobaeza: Yeah, but anyway, the issue is there and all of the v17 pretends to go that way: if using such system, it will contaminate all modules with such file, which it's far from ideal, and annoy a lot those that are not using the PIP current.

ivs-cetmix: > Use false here in order to remove that files

Done. Thank you for the hint!

pedrobaeza: The files are still there. They are a leftover.

ivs-cetmix: > The files are still there. They are a leftover.

Ooops.. should be fine now

pedrobaeza (approved):
/ocabot merge patch

attachment_zipped_download/pyproject.toml
sbidoul: @pedrobaeza @ivs-cetmix For 17, a pyproject.toml in each addon will replace the setup.py files about which some people were complaining too. They we got used to it. In the end the result will be much cleaner.

But we should not have them for 16.0 by default. Normaly this copier question is off by default, I check.

PR #446, opened 2023-12-21
17.0

m1k3lm:
/ocabot migration document_knowledge document_page


pedrobaeza:
Thanks for the contribution.

Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0.

If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance.

m1k3lm:
Thanks @pedrobaeza! fixed at https://github.com/OCA/knowledge/pull/447 once merged I'll create a separate PR for document_page module.

PR #448, opened 2024-01-09
16.0[IMP]document_page: menu_parent_id context domain of menu creation

Tisho99:
On the document_page module, there is one use case when we can't correctly create a menu item for a page.

This case occurs when we want to create the document menu under a new "Documentation" menu. Example:

Sales
    - Documentation
        - Page 1
        - Page 2
        ...

Steps to reproduce:

1. Create the "Documentation" parent menu, as it is a folder we dont fill the 'action' field.
2. Go to a knowledge document form, and select action -> create mene. You can see that the "Documentation" menu does not appear on the parent menu's list.

Captura de pantalla de 2024-01-09 13-59-25

The reason of the problem is that, by default, we only can see the menus that have an action, or a child with actions. 
To see all, odoo gives a context parameter that can be placed on the field: context="{'ir.ui.menu.full_list': True}".
However, with this context on the menu_parent_id of the document_page create menu wizard the document users are able to see menus they must not be able to see due to permission groups; and they can create aditional menus under the unauthorized ones.
To solve this, a new context parameter has been created: "ir.ui.menu.authorized_list", that allows to see all menus, except the unauthorized by permission groups

PD: A unit test is in progress

### DIFF
--- a/document_page/README.rst
+++ b/document_page/README.rst
@@ -93,6 +93,7 @@ Trobz
 * `Sygel <https://www.sygel.es>`_:
 
   * Ángel García de la Chica Herrera
+  * Alberto Martínez Rodríguez
 
 Other credits
 ~~~~~~~~~~~~~
--- a/document_page/models/__init__.py
+++ b/document_page/models/__init__.py
@@ -2,3 +2,4 @@
 
 from . import document_page
 from . import document_page_history
+from . import ir_ui_menu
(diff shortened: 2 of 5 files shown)


Tisho99:
@ValentinVinagre

ValentinVinagre:
@HaraldPanten @etobella 😄

etobella (approved):

Tisho99:
@etobella 

I will put this PR in draft for some time because i am developing a unit test

Thank you for your review!

ValentinVinagre (approved):
LGTM 👍🏻

PR #465, opened 2024-03-07
[16.0][MIG] document_page_partner

ntsirintanis:
Renamed `document_page_partner_id` to `document_page_partner` and ported it from 10 to 16, with intermediate MR https://github.com/OCA/knowledge/pull/314


thomaspaulb:
Pretty straightforward - it's just the adding of 1 field. Why still draft? And there's 1 pre-commit issue

thomaspaulb:
@ntsirintanis Somehow the commits seem a bit strange - the pre-commit commit removes all the files, whereas the migration commit adds them all again under a new module name. It would be better if perhaps there's a separate "rename" commit, which renames the module, after which the pre-commit is run. Or, the renaming can be included in the migration to 16.0.

flotho (approved):
thanks
Expected behaviour

gaelTorrecillas (approved):

thomaspaulb:
@ntsirintanis Still, I don't understand the commit structure.

- `document_page_partner_id` is added
- `document_page_partner` is added (Danny's copy)
- `pre-commit` is run, which also includes removal of `document_page_partner_id`
- `document_page_partner` has some migration stuff

I understand how this came to be, because you based yourself on https://github.com/OCA/knowledge/pull/314, who made the original choice of not renaming but copying.

But in this way you lose commit history and you can't `blame` back anymore to the source of a commit.

If you're going to do it in this way, you might as well remove the commits related to `document_page_partner_id` completely from the history, because they carry no value anymore. But better of course is to git-commit the actual rename, so as to keep history. You could do something like:

```
cp -R document_page_partner /tmp
git reset HEAD~3
git mv document_page_partner_id document_page_partner
pre-commit run -A
git commit -A -m "[UPD] pre-commit"
rm -Rf document_page_partner
cp -R /tmp/document_page_partner .
git add .
pre-commit run
git commit -m "[MIG] document_page_partner to 16.0"
```

HaraldPanten (approved):

PR #475, opened 2024-04-08
[FIX] incorrect check for users to send notification

iTecan:
Error document_page_approval->document_page_history.py
line 81:
rec.message_subscribe([u.id for u in users]) # WRONG
rec.message_subscribe([u.id for u in users.partner_id]) # CORRECT

The incorrect version of the code checked the sending of the notification based on the user id, the problem is that the notifications should be checked against the partner_id as it may be different from the user_id and this sometimes sends messages to the wrong contacts.


fkantelberg:
FYI there was already a commit for it done in 13.0 and I forward ported it to the versions in the prs #480 #481 #479

iTecan:
@fkantelberg thanks for the info, I saw that your pull for 15.0 (I forgot to wrote it on the title my bad) is going to merge so I'm closing this pull

PR #285, opened 2021-02-23
[13.0][FIX] document_page_approval: Add correct followers in history pages

victoralmau:
Add correct followers in history pages

It's related to 12.0: https://github.com/OCA/knowledge/pull/284

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT28348

PR #481, opened 2024-04-25
[16.0][FIX] document_page_approval: Add correct followers in history pages

fkantelberg:
Forward port of #285 

Seems like this wasn't ported beyond 13.0.

Can you review @victoralmau @pedrobaeza and @joao-p-marques


pedrobaeza (approved):
/ocabot merge patch

PR #491, opened 2024-07-16
[16.0][FIX] document_page_tag: Use api.model_create_multi

anusriNPS:
Using decorator api.model_create_multi for create method in document_page_tag model in order to avoid below warning:
   
   WARNING devel odoo.api.create: The model odoo.addons.document_page_tag.models.document_page_tag is not overriding the create method in batch

### DIFF
--- a/document_page_tag/models/document_page_tag.py
+++ b/document_page_tag/models/document_page_tag.py
@@ -15,10 +15,13 @@ class DocumentPageTag(models.Model):
         ("unique_name", "unique(name)", "Tags must be unique"),
     ]
 
-    @api.model
-    def create(self, vals):
+    @api.model_create_multi
+    def create(self, vals_list):
         """Be nice when trying to create duplicates"""
-        existing = self.search([("name", "=ilike", vals["name"])], limit=1)
-        if existing:
-            return existing
-        return super().create(vals)
+        for vals in vals_list:
+            existing = self.search([("name", "=ilike", vals["name"])], limit=1)
+            if existing:
+                vals_list.remove(vals)
+                if not vals_list:
+                    return existing
+        return super().create(vals_list)
--- a/document_page_tag/tests/test_document_page_tag.py
+++ b/document_page_tag/tests/test_document_page_tag.py
@@ -8,15 +8,28 @@ from odoo.tools.misc import mute_logger
 
 class TestDocumentPageTag(TransactionCase):
     def test_document_page_tag(self):
-        testtag = self.env["document.page.tag"].name_create("test")
+        test_tag = self.env["document.page.tag"].name_create("test")
         # check we're charitable on duplicates
         self.assertEqual(
-            testtag,
+            test_tag,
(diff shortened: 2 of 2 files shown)


document_page_tag/models/document_page_tag.py
```diff
@@ -15,10 +15,11 @@ class DocumentPageTag(models.Model):
...
         """Be nice when trying to create duplicates"""
-        existing = self.search([("name", "=ilike", vals["name"])], limit=1)
-        if existing:
-            return existing
-        return super().create(vals)
+        for vals in vals_list:
+            existing = self.search([("name", "=ilike", vals["name"])], limit=1)
+            if existing:
+                return existing
+        return super().create(vals_list)
```
HekkiMelody: **Issue**: When trying to create two tags at the same time, one that exists and one that doesn't, we return the one tag that does exist and don't create the one that doesn't.

HekkiMelody (changes requested):

document_page_tag/models/document_page_tag.py
anusriNPS: Changes included to handle tag creation with one exist and not exist at the same time

document_page_tag/tests/test_document_page_tag.py
```diff
@@ -14,9 +14,21 @@ def test_document_page_tag(self):
...
         )
+
+        # check multiple record creation
+        testtag1 = self.env["document.page.tag"].create(
+            [{"name": "test1"}, {"name": "test"}, {"name": "test2"}]
+        )
+        self.assertEqual(len(testtag1), 2)
+        self.assertEqual(testtag1[0].name, "test1")
+        self.assertEqual(testtag1[1].name, "test2")
+        testtag2 = self.env["document.page.tag"].create([{"name": "test"}])
```
PicchiSeba: **nitpick**: separate test2 section from test1's

PicchiSeba (commented):
Code Review: generally seems good. The only issue I have is with the variable names in the test (not your fault, they were already done like this).
We might consider converting them to snake_case format

document_page_tag/tests/test_document_page_tag.py
anusriNPS: Done

anusriNPS: Changed variable name to snakecase format

PicchiSeba:
watch out for the pre-commit error at line 68 in `document_page_tag/tests/test_document_page_tag.py`

document_page_tag/tests/test_document_page_tag.py
anusriNPS: Missed to  commit pre commit changes after pushing, updated the required changes

PicchiSeba (approved):
Code review: LGTM

HekkiMelody (approved):

dreispt (approved):

PR #492, opened 2024-07-24
[16.0][IMP] document_page_access_group*: Users compatibility

victoralmau:
Users compatibility

Please @pedrobaeza and @carolinafernandez-tecnativa can you review it?

@Tecnativa tt48786

### DIFF
--- a/document_page_access_group/models/document_page.py
+++ b/document_page_access_group/models/document_page.py
@@ -1,10 +1,19 @@
 # Copyright 2022 Manuel Regidor <manuel.regidor@sygel.es>
+# Copyright 2024 Tecnativa - Víctor Martínez
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import UserError
 
 
 class DocumentPage(models.Model):
     _inherit = "document.page"
 
     groups_id = fields.Many2many(comodel_name="res.groups", string="Groups")
+    user_ids = fields.Many2many(comodel_name="res.users", string="Users")
+
+    @api.constrains("groups_id", "user_ids")
+    def check_document_page_groups_users(self):
+        for _item in self.filtered(lambda x: x.groups_id and x.user_ids):
+            raise UserError(_("You cannot set groups and users at the same time."))
+        return True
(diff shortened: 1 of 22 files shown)


victoralmau:
Ping @pedrobaeza and @carolinafernandez-tecnativa

document_page_access_group/models/document_page.py
```diff
@@ -1,10 +1,19 @@
...
 class DocumentPage(models.Model):
     _inherit = "document.page"
 
     groups_id = fields.Many2many(comodel_name="res.groups", string="Groups")
+    user_ids = fields.Many2many(comodel_name="res.users", string="Users")
+
+    @api.constrains("groups_id", "user_ids")
+    def check_document_page_groups_users(self):
+        for _item in self.filtered(lambda x: x.groups_id and x.user_ids):
+            raise UserError(_("You cannot set Groups and Users."))
```
pedrobaeza:
```suggestion
            raise UserError(_("You cannot set groups and users at the same time."))
```

document_page_access_group_user_role/migrations/16.0.1.1.0/post-migration.py
```diff
@@ -0,0 +1,12 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    """Pages that had roles should now have the correct users."""
+    pages = env["document.page"].sudo().search([("role_ids", "!=", False)])
+    for page in pages:
+        page.role_ids = False
```
pedrobaeza, line 12: Why emptying roles?

victoralmau: Ok sorry, it's already changed. You get the users that you will have to have linked (to define them later) and then you empty the groups. This is to avoid the constrains _check_document_page_groups_users_

document_page_access_group/models/document_page.py
```diff
@@ -1,10 +1,19 @@
...
 
 
 class DocumentPage(models.Model):
     _inherit = "document.page"
 
     groups_id = fields.Many2many(comodel_name="res.groups", string="Groups")
+    user_ids = fields.Many2many(comodel_name="res.users", string="Users")
+
+    @api.constrains("groups_id", "user_ids")
+    def check_document_page_groups_users(self):
```
carolinafernandez-tecnativa, line 16: This constrains is not being applied, check this example:

1. Add one role to one document page
image

2. Add one groups and save:

image

3. Roles now gets invisible
image

4. Remove group and then you will see that role is still applied:

image

victoralmau: I think it is important to clarify that constrains is related to **groups** and **users** and is something that cannot be achieved by UX. If we perform a `write()` of a page set groups and users the error will be displayed (this is what we want to avoid, the data inconsistency).

carolinafernandez-tecnativa (approved):
LGTM Thanks :)

pedrobaeza (approved):
/ocabot merge nobump

PR #503, opened 2024-11-05
[18.0][MIG] document_knowledge: Migration to 18.0

Kimkhoi3010:


jdoutreloux (commented):
It might be nice to set the `path` attribute on the `knowledge_action_documents` to have a user-friendly url. 
But otherwise lgtm

document_knowledge/views/document_knowledge.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="knowledge_action_documents" model="ir.actions.act_window">
+        <field name="name">Documents</field>
```
jdoutreloux, line 4: `<field name="path">documents</field>`

Kimkhoi3010: @jdoutreloux thanks for your time!!

TDu (approved):

jdoutreloux (approved):

houssine78 (approved):
LGTM

houssine78:
@pedrobaeza a new PR that can be merge here :)

document_knowledge/models/res_config.py
```diff
@@ -9,25 +9,17 @@ class DocumentKnowledgeConfigSettings(models.TransientModel):
 
     module_attachment_indexation = fields.Boolean(
         "Attachments List and Document Indexation",
-        help="Document indexation, full text search of attachements.\n"
```
pedrobaeza: I prefer that you remove all the options, as this is not valid if the rest of the modules aren't there.

document_knowledge/readme/CREDITS.md
```diff
@@ -1,3 +1,8 @@
 The development of this module has been financially supported by:
 
 - Camptocamp
+
+The migration of this module from 17.0 to 18.0 was financially supported
```
pedrobaeza: Remove the previous text.

pedrobaeza (changes requested):
Controversial opinion about removing the settings part, but this is a frequent source of questions/problems cc @max3903

document_knowledge/models/res_config.py
Kimkhoi3010: @pedrobaeza Thanks for your time!!
Can you explain more so I can better understand what you want.

pedrobaeza: I mean that having these `module_*` fields is for installing modules that are still not migrated, but even if migrated, you may not have those modules present on your Odoo instance, so clicking on the checks provoke confusion and misleading in users, so I propose to remove these fields from the configuration.

pedrobaeza:
Thanks for the changes. I think you have stripped now too much, as I was just talking about `module_*` fields, not the rest.

Kimkhoi3010:
> Thanks for the changes. I think you have stripped now too much, as I was just talking about `module_*` fields, not the rest.

It was my fault. I fixed it. Thanks for your time!!

houssine78:
thanks @pedrobaeza for this insightful review.

pedrobaeza (approved):
/ocabot merge nobump

PR #506, opened 2024-11-07
[17.0][FIX] document_url: Fix Download attachment component

maq-adhoc:
This commit (https://github.com/odoo/odoo/commit/366676cafdce00d55823c6daf41452b0c2373e4d) move download buttons into a sub-component. 
I modified the module behavior to use canDownload() method instead of adding a t-if in the view and prevent the chatter component from breaking when the message has attachments.

### DIFF
--- a/document_url/static/src/js/url.esm.js
+++ b/document_url/static/src/js/url.esm.js
@@ -52,6 +52,11 @@ patch(AttachmentList.prototype, {
      *
      * @returns {String}
      */
+    canDownload(attachment) {
+        return (
+            super.canDownload(attachment) && attachment.mimetype != "application/link"
+        );
+    },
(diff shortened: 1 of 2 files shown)


lef-adhoc (approved):
LGTM

giarve (approved):
I can confirm it needs a patch for 17.0 and this one is not looking bad

pedrobaeza:
Please use proper commit message:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

maq-adhoc:
@pedrobaeza I think it's okay now. Thanks

pedrobaeza (approved):
Thanks, that's it.

/ocabot merge patch

PR #512, opened 2024-12-18
[17.0][MIG] attachment_zipped_download: Migration to 17.0

victoralmau:
Migration to 17.0

Please @pedrobaeza and @chienandalu can you review it?

@Tecnativa TT52268


test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo_test_helper
```
pedrobaeza, line 1: Looking at the use of this library, it seems to me very weird, as `res.partner` is in base module, so I don't think we need to use `FakeModelLoader`. Can you please check about this for removing unneeded dependencies?

victoralmau: `FakeModelLoader` is still needed because even though `res.partner` is a base model, it is not extended by `ir.attachment.action_download` (because it is not needed); instead, `FakeModelLoader` is not needed in `product_attachment_zipped_download`.

pedrobaeza (approved):

acysos (approved):
Working in a production database. All ok.

PR #519, opened 2025-01-21
15.0[IMP] document_page: menu_parent_id context domain of menu creation

ValentinVinagre:
Port -> https://github.com/OCA/knowledge/pull/448

### DIFF
--- a/document_page/README.rst
+++ b/document_page/README.rst
@@ -93,6 +93,7 @@ Trobz
 * `Sygel <https://www.sygel.es>`_:
 
   * Ángel García de la Chica Herrera
+  * Alberto Martínez Rodríguez
 
 Other credits
 ~~~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


ValentinVinagre:
@HaraldPanten @etobella

HaraldPanten (changes requested):
Functional review. But you need to remove the Odoo version in your commit message

ValentinVinagre:
@HaraldPanten is the same name of the commit in v.16 . it's a port.

HaraldPanten:
It's a port, but it has an incorrect name. Enric, we should correct the commit message, right?

ValentinVinagre:
> It's a port, but it has an incorrect name. Enric, we should correct the commit message, right?

It would have to be fixed in v16 first and then the fix brought to v15

etobella:
I would leave the message as it is. As a good practice, it is better to avoid version name on the commits. The only exception is migration commits

HaraldPanten (approved):

PR #524, opened 2025-02-17
[18.0][MIG] document_page_approval: Migration to 18.0

BhaveshHeliconia:


dreispt:
Can you please rebase on latest 18.0 - that should fix the tests.

BhaveshHeliconia:
@dreispt, Please Review. It's done.

document_page_approval/models/document_page_history.py
```diff
@@ -151,8 +151,9 @@ def _compute_page_url(self):
             )
 
             page.page_url = (
-                "{}/web#db={}&id={}&" "model=document.page.history"
-            ).format(base_url, self.env.cr.dbname, page.id)
+                f"{base_url}/web#db={self.env.cr.dbname}&id={page.id}&"
+                "model=document.page.history"
```
dreispt, line 158: Nota that these URLs are still supported, but have changed and might stop working in future versions.

dreispt (approved):

document_page_approval/models/document_page_history.py
```diff
@@ -156,14 +160,18 @@ def _compute_page_url(self):
...
             if prev:
-                rec.diff = self._get_diff(prev.id, rec.id)
+                raw_diff = self._get_diff(prev.id, rec.id)
             else:
-                rec.diff = self._get_diff(False, rec.id)
+                raw_diff = self._get_diff(False, rec.id)
+
+            cleaned_diff = re.sub(r'data-last-history-steps=".*?"', "", raw_diff)
+
+            rec.diff = cleaned_diff
```
victoralmau: What is the reason for this? Sorry, but I don't understand (if it is necessary, perhaps it would be appropriate to explain in the method the reasons, so that we all know it in the future).

document_page_approval/tests/test_document_page_approval.py
```diff
@@ -9,18 +11,39 @@ def setUpClass(cls):
...
+                "perm_write": True,
+                "perm_create": True,
+                "perm_unlink": True,
+            }
+        )
+
+        # Disable restrictive record rules for the test
+        cls.env["ir.rule"].search(
+            [("model_id", "=", cls.env["ir.model"]._get_id("document.page.history"))]
+        ).write({"active": False})
```
victoralmau: I don't understand why these changes are necessary. In migrations, the proper thing to do is to reduce the diff (for better readability).

document_page_approval/tests/test_document_page_approval.py
```diff
@@ -0,0 +1,275 @@
...
+        cls.page1 = cls.env.ref("document_page.demo_page1")
+
+        # Create test user without groups first
+        cls.user2 = new_test_user(cls.env, login="test-user2", groups="base.group_user")
+
+        # Ensure user2 has the approver group
+        cls.approver_gid = cls.env.ref(
+            "document_page_approval.group_document_approver_user"
+        )
+        cls.user2.write({"groups_id": [Command.link(cls.approver_gid.id)]})
```
victoralmau:
```suggestion
        cls.user2 = new_test_user(cls.env, login="test-user2", groups="document_page_approval.group_document_approver_user")

        cls.approver_gid = cls.env.ref(
            "document_page_approval.group_document_approver_user"
        )
```
The `base.group_user` permission is inherited indirectly with `document_knowledge`

victoralmau (changes requested):

BhaveshHeliconia:
@victoralmau, Please Review.it's done.

document_page_approval/tests/test_document_page_approval.py
BhaveshHeliconia: we have to add this because of apply write access to document.page.history

victoralmau: No, doing this in a test is not necessary for at least **99%**, and in this case it is not. Doing this indicates that something defined in the test is incorrect. Checking in detail I have already detected what happens, I suggest it now (remember, a test should test a real case, and there was some unnoticed but important step that was not being done).

document_page_approval/tests/test_document_page_approval.py
```diff
@@ -103,41 +127,153 @@ def test_change_request_from_scratch(self):
         )
 
         self.assertEqual(chreq.state, "draft")
-        self.assertNotEqual(page.content, chreq.content)
-        self.assertNotEqual(page.approved_date, chreq.approved_date)
-        self.assertNotEqual(page.approved_uid, chreq.approved_uid)
 
-        chreq.action_to_approve()
```
victoralmau: It is not necessary to remove it, it is important.

document_page_approval/tests/test_document_page_approval.py
```diff
@@ -103,41 +127,153 @@ def test_change_request_from_scratch(self):
...
+        # Move change request to 'to approve' state
+        chreq.with_user(self.user2).action_to_approve()
```
victoralmau: This is not necessary, it is already in “to approve” status requested by the user who created the history.

document_page_approval/tests/test_document_page_approval.py
```diff
@@ -103,41 +127,153 @@ def test_change_request_from_scratch(self):
...
-        chreq.action_draft()
+        chreq.with_user(self.user2).action_draft()
         self.assertEqual(chreq.state, "draft")
-        self.assertNotEqual(page.content, chreq.content)
-        self.assertNotEqual(page.approved_date, chreq.approved_date)
-        self.assertNotEqual(page.approved_uid, chreq.approved_uid)
 
-        chreq.action_approve()
+        # Approve the change request
+        chreq.with_user(self.user2).action_approve()
```
victoralmau, line 124: If you want `user2` to approve it, you need it to be in “to approve” status, so you will need to do the following first

```
chreq.action_to_approve()
self.assertEqual(chreq.state, "to approve")
```

victoralmau (changes requested):
Now you can remove the code from https://github.com/OCA/knowledge/pull/524#discussion_r2012509984

victoralmau:
Please, cherry-pick https://github.com/OCA/knowledge/pull/531 to commit history before migration commit.

BhaveshHeliconia:
@victoralmau, Please Review.i updated test cases.

victoralmau (approved):
Thank you, good job.

Remember for the future that it is not necessary to add line breaks. I comment this because it helps to reduce the diff of the migration commit and is more readable.

PR #534, opened 2025-04-09
[FIX] document_page fr translation

youelEl:


pedrobaeza:
Please use the proper tool for doing translations:

https://odoo-community.org/resources/translate

PR #536, opened 2025-04-15
[17.0][FIX] knowledge icon

Gerloav:


pedrobaeza:
Isn't that icon used in other module?

Gerloav:
knowledge module

pedrobaeza:
Yes, so you can't copy something of enterprise.

pedrobaeza:
I'm afraid changing colors is not enough. We should provide an original one.

Issue #502, opened 2024-11-04
Duplicated functionality


etobella:
I found that there are 2 modules in the repository that provide the same functionality:

document_page_access_group: Since v14.0. Done by @manuelregidor @Sygel-technology
document_page_group: Since v11.0 Done by @etobella @dixmit

We should need to merge them on a single one. I don't care if we take document_page_access_group because it has some extra functionalities.

My proposal, would be to do the merge on v18 and keep all the authors

pedrobaeza:
cc @victoralmau

victoralmau:
I understand the confusion with both module and personally I think they should be merged in v18 in 1 or even in `document_page` (with different options at configuration level).

I think it is interesting to clarify how each module works.

**document_page_group**
"_Visible to_” is set (`direct_group_ids`) which allows to select groups (sales user for example). The `group_ids` field is also set with a compute.

A user will have access to a page if he/she does NOT have any group set and/or is part of one of them.

**document_page_access_group**
"_Groups_” (`groups_id`) and "_Users_" (`user_ids`) are set, neither of them is compute.
It is possible to set either Groups or Users (not both) on a page.

Several permissions are created: 
- Document Page Access: The user will be able to view pages that do not have “Groups” and “Users” or if the user is one of the indicated or if the user is part of one of the indicated groups.
- Document Page Full Access: Has access to all pages.

manuelregidor:
@ValentinVinagre What do you think?

etobella:
Well, the module has been migrated to 18... 

@ValentinVinagre your comment would be appreciated, if you find it right, I can make the PR on OpenUpgrade and add the authors on the module.

ValentinVinagre:
@etobella @manuelregidor @victoralmau 
Hi, sorry, I just read the issue.
I don't see any problem with merging them, as long as the functionality is correct.

pedrobaeza:
Closing as the PRs have been done.

PR #546, opened 2025-05-13
[18.0] [IMP] document_page_access_group: update contributors from document_page_group

luisDIXMIT:
Due to issue #502, document_page_group will not be continued, so the contributors can be maintained in the merged module


pedrobaeza (approved):
/ocabot merge patch

PR #550, opened 2025-06-23
[18.0][MIG] document_page_reference: Migration to 18.0

Anushapadinharepurakkal:


victoralmau (changes requested):
Can you review tests and fix it?

`WARNING odoo odoo.addons.document_page_reference.models.document_page: Render failed for 16: unexpected char '&' at 11 `

document_page_reference/models/document_page.py
```diff
@@ -0,0 +1,106 @@
...
+                references = re.findall(r"\$\{([^}]+)\}", raw)
+                for ref in references:
+                    if not re.match(r"^[a-zA-Z_][a-zA-Z0-9_]*$", ref):
+                        raise ValidationError(_("Invalid reference: %s") % ref)
+                converted = re.sub(r"\$\{([\w_]+)\}", r"{{ resolve('\1') }}", raw)
+                template = env.from_string(converted)
+                rendered = template.render(resolve=record._resolve_reference)
+                record.content_parsed = rendered
+            except Exception as e:
+                _logger.warning("Render failed for %s: %s", record.id, e)
```
victoralmau:
```suggestion
                _logger.info("Render failed for %s: %s", record.id, e)
```
This change is necessary for the CI to be green.

victoralmau (changes requested):
The extra commit "_[FIX]document_page_reference: render issue fixed_" is not necessary, those changes must be added in the migration commit.

document_page_reference/views/document_page.xml
```diff
@@ -0,0 +1,70 @@
...
+            <field name="content" position="before">
+                <field
+                    name="content_parsed"
+                    class="oe_read_only"
+                    widget="document_page_reference"
+                />
+            </field>
+        </field>
+    </record>
+
```
victoralmau, line 30:
```suggestion
```

document_page_reference/views/document_page.xml
```diff
@@ -0,0 +1,70 @@
...
```
victoralmau, line 48:
```suggestion
```

document_page_reference/views/document_page.xml
```diff
@@ -0,0 +1,70 @@
...
+        <field name="name">document.page.search (in knowledge_reference)</field>
+        <field name="model">document.page</field>
+        <field name="inherit_id" ref="document_page.view_wiki_filter" />
+        <field name="arch" type="xml">
+            <field name="parent_id" position="before">
+                <field name="reference" />
+            </field>
+        </field>
+    </record>
+
```
victoralmau, line 59:
```suggestion
```

document_page_reference/tests/test_document_reference.py
```diff
@@ -0,0 +1,65 @@
+# Copyright 2019 Creu Blanca
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
+
+
+class TestDocumentReference(TransactionCase):
```
victoralmau, line 8:
```suggestion
from odoo.addons.base.tests.common import BaseCommon


class TestDocumentReference(BaseCommon):
```

victoralmau (approved):
Code review OK.

Some non-blocking suggestions.

marcelsavegnago (approved):

victoralmau:
Ping @pedrobaeza

victoralmau:
Functionally reviewing this I see that there are many things wrong:

1- Text to replace
Currently as stated in USAGE you should use `${XXX}` but I think it should be changed to `{{XXX}}`.

2- Replacement does not work:
It does not do the proper replacement in https://github.com/OCA/knowledge/blob/18.0/document_page_reference/models/document_page.py#L67 because ' is defined `&#39;` instead of `'`, which causes https://github.com/OCA/knowledge/blob/18.0/document_page_reference/models/document_page.py#L68 not to be done properly. This will also cause problems when there is actual html content because the `from_string()` method only supports a string, nothing else. I think this “explains” why some tests https://github.com/OCA/knowledge/commit/7307e198badcc4f0722c15982b459799a1d1902e#diff-d117284773ecae79e7460e8fee3edaeb414c6040e974716da04277523354bd1eL35 were removed.

IMO I think you should remove everything related to jinja2 and simplify it a lot, do a regex to find all occurrences of `{{xxx}}` and do the replacement directly by https://github.com/OCA/knowledge/blob/1d99ddd812232425b627a3c82cb264018d44e468/document_page_reference/models/document_page.py#L7.

I really think that this module adds too much complexity, because perhaps the write/inverse method of the content should be intercepted so that when it is modified, the content is re-defined (or a reference is modified) and already has the “replace” applied (https://github.com/OCA/knowledge/blob/1d99ddd812232425b627a3c82cb264018d44e468/document_page_reference/models/document_page.py#L76). This way it would not be necessary to do anything in https://github.com/OCA/knowledge/blob/18.0/document_page_reference/views/report_document_page.xml#L5.

3- .js code does not work

I think I understand what is intended in https://github.com/OCA/knowledge/blob/18.0/document_page_reference/static/src/js/editor.esm.js but it does not work. With the final approach of the previous point, this would not be necessary because when saving I would already be replacing what corresponds.

4- Change t-raw.
Although it is not directly related to this module, you could change https://github.com/OCA/knowledge/blob/18.0/document_page/views/report_document_page.xml#L7 t-raw by t-out to avoid the warning log.

PR #560, opened 2025-09-30
[18.0][IMP] document_knowledge: Improve logo

etobella:
Previously:

image

Now:

image

I used a bulblight as it was easier, trying to put there a braing might be a lot harder.


pedrobaeza:
I think the light is used for another concepts like `idea` and similar, so it would be good to keep the topic.

etobella:
image

WDYT?

pedrobaeza (approved):

leyes-franco (approved):

PR #561, opened 2025-10-15
[I18N] Translate Vietnamese

tiennq1408:
Update and Add vi.po


etobella:
You need to use translation.odoo-community.org for that. If you don't have access, send an email to transbot@odoo-community.org

For more information: https://odoo-community.org/resources/translate

PR #563, opened 2025-10-27
[18.0][IMP] document_page: Use odoo standard diff tool for showing the real difference

etobella:
Shows differences by markdown:

image


pedrobaeza:
And what happens with HTML?

etobella:
The proble with diff made with HTML is that is usually less clear to final users. It includes a lot of changes, that are not clear for them. Markdown is clear for them. In any case, with the current change, the mardown difference is only accessible if you have defined it (we could even do a parameter for this in order to make it cleaner and keep HTML by default).

pedrobaeza:
Yeah, what I mean is that right now, there's no HTML diff, but plain text over HTML code diff, so what I'm talking about is to render the HTML and the marks of diff.

etobella:
I am not sure if that is factible... for example Github is not doing that with html files It shows the diff of lines.

pedrobaeza:
Well, but that's because it's mostly code, but here we have documentation in rich text format, so I think it's a different beast, and we should provide proper tools for it.

luisDIXMIT (approved):
Tested locally and LGTM!

etobella:
@pedrobaeza There is something like: https://pypi.org/project/html-diff/. What do you think?

pedrobaeza:
Yeah, it seems interesting.

etobella:
@pedrobaeza Something like this?

image

pedrobaeza:
Yeah, it looks great!!

LoisRForgeFlow (approved):
Thanks!! for sure is better than it is now... I got very used to ignore the diff because it was unusable...

luisDIXMIT (approved):
Retested locally after recent changes and it works fine. Good job!

luisDIXMIT (commented):
I detected something weird. After deleting one line, the whole document is crossed out and detected as new.


https://github.com/user-attachments/assets/0f514c5d-a944-4a0f-b886-3d9ae6fc7458

etobella:
@luisDIXMIT I found the same at some point, but I think this is a consequence of the import of the demo data.

Can you try with a document created by you directly?

luisDIXMIT:
> @luisDIXMIT I found the same at some point, but I think this is a consequence of the import of the demo data.
> 
> Can you try with a document created by you directly?

I created a new test document with one line. After saving it, I added another line, and this is how it was detected:

image

etobella:
I found the problem and a much better solution.

Odoo provides a diff method, so we can do everything without external libraries.

etobella:
I think it is much cleaner now:

image


@pedrobaeza WDYT?

It can be ported automatically to v17, but on 16 we might need to add some code from odoo on it

- https://github.com/odoo/odoo/blob/c4e21ef6ef835d3af02ca9cee495d588a3cc8c72/addons/html_editor/models/diff_utils.py#L30

luisDIXMIT (approved):
Tested with the latest changes and LGTM!

pedrobaeza (approved):
Thanks for the great improvement.

/ocabot merge minor

PR #580, opened 2025-12-31
[16.0][IMP] document_page: Adapt demo to new format

etobella:
I found out that the demo data was giving a lot of problems due to the static data used, as the Html fields have changed a lot.

This should fix most of the problems with the data (there are some weird cases, but less than before).


pedrobaeza (approved):
Let's fix them!

/ocabot merge patch

PR #581, opened 2026-01-08
[18.0][FIX] document_url: Easier template inheriting

ypapouin:
- In a first step, instead of replacing a node content with its verbatim copy, we uses the xpath "replace" command with "$0" magic keyword to move all the \<t\> content into a \<div\>.
- In a second step, the "Add URL" button is added after the existing "Attach files"button.


pedrobaeza (approved):
Tested and working. Thanks for the improvement.

marcelsavegnago (approved):
LGTM

PR #584, opened 2026-03-02
[18.0][FIX] document_page_approval: Avoid errors when creating a page history and send mail

victoralmau:
Avoid errors when creating a page history and end mail

Example use case:
- Modify the Odoo Features category, check Require approval, and set Approver group > Administration / Access Rights
- Modify the Marc Demo user and define Document Knowledge > Editor permissions
- Go to the Odoo Features category and, taking into account the ID, delete the mail_message record table with the same ID from the database
- With the Marc Demo user, go to the Knowledge > Odoo Features menu and enter the Odoo 15.0 Functional Demo page (it is important to follow these specific steps)
- Modify the page and save

Before this change, an error could occur because the action https://github.com/OCA/knowledge/blob/0881d75a35e7fc2f814110df4878177e805e589b/document_page/views/document_page_category. xml#L83C17-L83C42 defined a default_parent_id=x (page category ID) if there was no mail_message linked to that category ID; now the value of default_parent_id is removed to avoid this specific error.

Please @pedrobaeza and @pilarvargas-tecnativa can you review it?

@Tecnativa TT61290

### DIFF
--- a/document_page_approval/models/document_page_history.py
+++ b/document_page_approval/models/document_page_history.py
@@ -4,6 +4,7 @@
 
 from odoo import fields, models
 from odoo.exceptions import UserError
+from odoo.tools.misc import clean_context
 from odoo.tools.translate import _
 
 
@@ -80,7 +81,10 @@ class DocumentPageHistory(models.Model):
                     [("groups_id", "in", guids), ("groups_id", "in", approver_gid.id)]
                 )
                 rec.message_subscribe(partner_ids=users.mapped("partner_id").ids)
-                rec.message_post_with_source(template)
+                # pylint: disable=W8121
+                rec.with_context(
+                    clean_context(self.env.context)
+                ).message_post_with_source(template)
             else:
                 # auto-approve if approval is not required
                 rec.action_approve()
--- a/document_page_approval/tests/test_document_page_approval.py
+++ b/document_page_approval/tests/test_document_page_approval.py
@@ -1,6 +1,7 @@
 from odoo import Command
 from odoo.exceptions import UserError
 from odoo.tests import new_test_user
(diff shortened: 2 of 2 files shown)


pedrobaeza (approved):

christian-ramos-tecnativa:
Can you be a bit more specific about the third step?

> Go to the Odoo Features category and, taking into account the ID, delete the mail_message record table with the same ID from the database

I'm not able to reproduce the error and assuming is because of this step. I went to mail_message and deleted the one linked to the Category `Odoo Features` the one saying "Document Page created", is that correct?

victoralmau:
> Can you be a bit more specific about the third step?
> 
> > Go to the Odoo Features category and, taking into account the ID, delete the mail_message record table with the same ID from the database
> 
> I'm not able to reproduce the error and assuming is because of this step. I went to mail_message and deleted the one linked to the Category `Odoo Features` the one saying "Document Page created", is that correct?

I think an example is the best way to explain it.

**Before**:
ejemplo-antes

**After**:
ejemplo-despues

document_page_approval/models/document_page_history.py
christian-ramos-tecnativa: As you said the problem clearing the hole context is that after the message is sent the user no longer has the default_parent_id even though he may think it is. WDYT about just clearing the context in the message_post? This way if the user create a new page it will keep the default category.
```suggestion
                rec.with_context(clean_context(self.env.context)).message_post_with_source(template)
```

christian-ramos-tecnativa (commented):
Thanks a lot for the Gif, I was able to replicate and understand the error

christian-ramos-tecnativa (approved):

PR #587, opened 2026-03-10
[16.0] copier update v1.39

CristianoMafraJunior:
Error pre-commit https://github.com/OCA/knowledge/actions/runs/22448036829/job/65008120137
cc @victoralmau @etobella


README.md
```diff
@@ -7,9 +10,13 @@
 
 <!-- /!\ do not modify above this line -->
 
+<<<<<<< before updating
```
etobella: I think this shouldn't be here 🤔

PR #600, opened 2026-03-20
[18.0][ADD] document_page_helpdesk_mgmt: add new module

CristianoMafraJunior:
#### Description
This PR introduces the document_page_helpdesk_mgmt module, enabling Document Page (Wiki) records to be linked to Helpdesk tickets (helpdesk.ticket).

cc @etobella @LoisRForgeFlow @victoralmau


etobella:
@CristianoMafraJunior First of all, thanks for the collaboration. However, I have some confusion on this topic.

IMO, tickets can be solved by using documentation (one document usually, but it could be more than one). The same on  the other side.

So, it should be a M2M, not a M2O. If it needed to be a M2O, for me the relation should be  1 Document has multiple Tickets.

Maybe you have a different picture and there are reasons for your schema. Can you enlighten us?

CristianoMafraJunior:
> @CristianoMafraJunior First of all, thanks for the collaboration. However, I have some confusion on this topic.
> 
> IMO, tickets can be solved by using documentation (one document usually, but it could be more than one). The same on the other side.
> 
> So, it should be a M2M, not a M2O. If it needed to be a M2O, for me the relation should be 1 Document has multiple Tickets.
> 
> Maybe you have a different picture and there are reasons for your schema. Can you enlighten us?

Changed the ticket-page relation from Many2one/One2many to Many2many to allow multiple wiki pages per ticket and reuse the same wiki page across multiple tickets.

PR #603, opened 2026-03-29
[18.0][attachment_preview] Add Office format preview via LibreOffice conversion

dnplkndll:
## Summary

Extends `attachment_preview` to support DOCX, XLSX, PPTX, DOC, XLS, PPT, and ODG preview in addition to the existing PDF + ODF support.

### How it works

- New HTTP endpoint: `GET /attachment_preview/office_to_pdf?model=&field=&id=&filename=`
- On request, converts the binary field content to PDF using LibreOffice headless (`libreoffice --headless --convert-to pdf`)
- Returns the PDF to ViewerJS for in-browser rendering
- Gracefully returns **HTTP 503** if LibreOffice is not installed — no preview button appears for Office files on deployments without LibreOffice

### Changes

| File | Change |
|------|--------|
| `controllers/main.py` | New LibreOffice conversion HTTP controller |
| `static/src/js/utils.esm.js` | `OFFICE_EXTENSIONS` added to `canPreview()`; `getUrl()` routes Office extensions through conversion endpoint with `?type=pdf` for ViewerJS |
| `static/src/js/web_views/fields/binary_field.esm.js` | Passes `filename` to `showPreview()` so conversion endpoint receives correct extension |

### Requirements

- LibreOffice must be installed in the Odoo environment for Office preview to function
- No new Python dependencies (uses stdlib `subprocess`, `tempfile`)
- Existing PDF + ODF preview unaffected

## Test plan

- [ ] Upload a `.docx` file to a record with a binary field → preview button (↗) appears
- [ ] Click preview → ViewerJS opens with converted PDF content
- [ ] Upload a `.xlsx` / `.pptx` → same behavior
- [ ] On an Odoo instance without LibreOffice → preview button does **not** appear for Office files (503 response causes button to stay hidden)
- [ ] Existing PDF preview unaffected
- [ ] Existing ODS/ODT preview unaffected

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pedrobaeza:
Why not integrating viewerjs directly into usual Odoo preview instead of having another preview mechanism like `attachment_preview`?

PR #609, opened 2026-04-22
[18.0][FIX] enable document reference link generation

KDRdev:
Currently referencing other documents with {{SOME_REF}} won't result in showing this as a link.

This PR addresses with issue and allows showing link after saving the document.

### DIFF
--- a/document_page_reference/tests/test_document_reference.py
+++ b/document_page_reference/tests/test_document_reference.py
@@ -18,7 +18,7 @@ class TestDocumentReference(BaseCommon):
             {"name": "Test Page 1", "content": Markup("{{r2}}"), "reference": "R1"}
         )
         cls.page2 = cls.page_obj.create(
-            {"name": "Test Page 1", "content": Markup("{{r1}}"), "reference": "r2"}
+            {"name": "Test Page 2", "content": Markup("{{r1}}"), "reference": "r2"}
         )
 
     def test_constraints_duplicate_reference(self):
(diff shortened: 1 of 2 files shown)


etobella:
TEsts are failing and use a proper commit description

KDRdev:
@etobella Tests have been fixed

document_page_reference/tests/test_document_reference.py
```diff
@@ -71,5 +72,15 @@ def test_get_formview_action(self):
             self.assertEqual(res.get(key), expected_value, f"Mismatch in key: {key}")
 
     def test_compute_content_parsed(self):
-        self.page1.content = Markup("<p></p>")
-        self.assertEqual(self.page1.content_parsed, Markup("<p></p>"))
+        self.page1.content = Markup("<p>{{r2}}</p>")
+        self.page1._compute_content_parsed()
```
etobella: You shouldn't need to execute compute functions

document_page_reference/tests/test_document_reference.py
```diff
@@ -4,11 +4,10 @@
 from markupsafe import Markup
 
 from odoo.exceptions import ValidationError
+from odoo.tests.common import TransactionCase
 
-from odoo.addons.base.tests.common import BaseCommon
 
-
-class TestDocumentReference(BaseCommon):
+class TestDocumentReference(TransactionCase):
```
etobella: Why did you change it back to Transactioncase? wasn't BaseCommon working?

KDRdev: Yes, I was getting not-null constraint error: `null value in column "autopost_bills" of relation "res_partner" violates not-null constraint` when using BaseCommon. With TransactionCase it works.

document_page_reference/tests/test_document_reference.py
KDRdev: You're right. Will remove unnecessary lines

document_page_reference/tests/test_document_reference.py
etobella: That probably happens because you installed several modules and this was providing a problem with that. LEave it as it was please :pray:

document_page_reference/tests/test_document_reference.py
KDRdev: Done

document_page_reference/tests/test_document_reference.py
KDRdev: Changed this back to BaseCommon. Thank you

etobella (approved):
Technically looks ok

jans23:
I tested the latest commit successfully.

PR #612, opened 2026-05-20
[FIX] document_page: make content field span full width in form views

miquelrosell99:
The content field was placed directly in the sheet without a group wrapper, which caused it to render in a narrow single column. Wrapping it in a <group colspan="2" nolabel="1"> makes it use the full available width in both the main document form and the history form.

### DIFF
--- a/document_page/views/document_page.xml
+++ b/document_page/views/document_page.xml
@@ -50,13 +50,17 @@
                             <field name="name" placeholder="Name" />
                         </h1>
                     </div>
-                    <field
-                        name="content"
-                        widget="html"
-                        placeholder="e.g. Once upon a time..."
-                        required="1"
-                        options="{'safe': True, 'codeview': True, 'collaborative': True}"
-                    />
+                    <group colspan="2" nolabel="1">
+                        <field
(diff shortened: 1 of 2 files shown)


pedrobaeza:
Can you put screenshots of before/after?

miquelrosell99:
By default it looked fine but in some cases for custom modules (I use one to use ckeditor, for example), it switched to one column layout. See example:
imagen

After my fix everything looks normal, even with custom modules that replace the html widget. And also, to my understanding, having fields in a group it is the standard approach to add fields in form view in Odoo?

pedrobaeza:
Uhm, then it seems a problem of the CK editor itself, not this one.

miquelrosell99:
Closing this PR as the issue is not in document_page but in the custom web_editor_ckeditor widget styling. The module document_page_form_content_layout is not needed; the root cause (missing block display and iframe box-sizing in the CKEditor widget) has been fixed in the custom web_editor_ckeditor module.

pedrobaeza:
Thanks for the feedback and glad that you fix it.

PR #613, opened 2026-05-20
[FIX] document_page_tag: convert create to api.model_create_multi

miquelrosell99:
Odoo 17 recommends batch-compatible create methods for better performance when creating records in bulk.

### DIFF
--- a/document_page_tag/models/document_page_tag.py
+++ b/document_page_tag/models/document_page_tag.py
@@ -15,10 +15,14 @@ class DocumentPageTag(models.Model):
         ("unique_name", "unique(name)", "Tags must be unique"),
     ]
 
-    @api.model
-    def create(self, vals):
+    @api.model_create_multi
+    def create(self, vals_list):
         """Be nice when trying to create duplicates"""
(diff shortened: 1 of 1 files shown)


CristianoMafraJunior (approved):

document_page_reference/tests/test_document_reference.py
```diff
@@ -29,7 +29,7 @@ def test_constraints_duplicate_reference(self):
     def test_constraints_invalid_reference(self):
         """Should raise if reference does not match the required pattern."""
         with self.assertRaises(ValidationError):
-            self.page2.write({"reference": self.page2.reference + "-02"})
+            self.page2.write({"reference": self.page2.reference + "@02"})
```
etobella: This change shouldn't be here?

miquelrosell99:
Hi @etobella, thanks for spotting that.

The change to `document_page_reference/tests/test_document_reference.py` was unrelated to this PR (it came from a separate experiment to allow dashes in references). I have rebased the branch on 17.0 and removed that commit, so the PR now only contains the `document_page_tag` `api.model_create_multi` conversion.

Please re-review when you have a moment.

PR #631, opened 2026-07-15
[18.0][IMP] document_page_reference: split linking feature into document_page_reference_link

miquelrosell99:
## Motivation

`document_page_reference` currently bundles two independent concerns:

1. the `reference` field on `document.page` (validation, uniqueness, auto-generation from the name, views);
2. the linking feature built on top of it: `content_parsed`, `{{reference}}` marker resolution into direct links, the report override, and the backend editor JS.

Some deployments want the field **without** the linking feature — either because they don't need it, or because they prefer a different linking implementation. Today that is not possible without uninstalling the whole module.

## Changes

- **`document_page_reference`** (18.0.3.0.0): keeps the `reference` field, its validation/auto-generation, and the form/search/list views. All linking code is removed.
- **`document_page_reference_link`** (new): contains everything that was moved out — `content_parsed`, `{{ref}}` resolution, report override, editor JS — with **no behavior changes**. Depends on `document_page_reference`.

All moved code (model methods, views, JS, tests) is byte-identical to the original; the diff is a pure code motion split.

## Upgrade path

- Updating `document_page_reference` alone **removes the linking feature** (the stored `content_parsed` column is dropped). To keep current behavior, install `document_page_reference_link` — the column is recreated and recomputed automatically on install.
- The `reference` field itself is unaffected (same model, same name): no data migration needed.

## Notes

- Nothing else in this repo depends on `document_page_reference` or its view XML IDs, so the split is safe repo-internally.
- I kept @etobella as maintainer of the new module and added myself as co-maintainer — happy to adjust.


etobella:
@miquelrosell99 I understand your need, however, the changes required to add the link are minimal and doesn't affect to first module, so it has sense to use them together, at the end, the user can decide not to make links, isn't it?

PR #635, opened 2026-08-12
[17.0][ADD] attachment_preview_office: preview Office attachments via LibreOffice conversion

alan196:
## Summary

New module that previews office documents (DOCX, XLSX, PPTX, DOC, XLS, PPT, ODT, ODS, ODP, ODG) in the browser — including chatter attachments — by converting them to PDF **server-side** with LibreOffice headless and rendering the result with Odoo's native PDF.js viewer.

Nothing leaves the server: no external viewer service (Microsoft Office Online, Google Docs) is involved, which makes it suitable for sensitive documents.

### Context

- `attachment_preview` was never migrated to 17.0 and its ViewerJS-based approach only covers PDF + ODF anyway.
- #603 adds Office-format preview to `attachment_preview` on 18.0 via LibreOffice conversion. This module brings the same conversion-endpoint logic to 17.0, but as a small standalone module: on 17.0 the native FileViewer already renders PDFs, so no ViewerJS or base module is needed.

### How it works

- HTTP endpoint `GET /attachment_preview_office/to_pdf?model=&field=&id=&filename=` converts the binary field content to PDF with `libreoffice --headless --convert-to pdf` (isolated `UserInstallation` profile per conversion, checksum-keyed disk cache under the filestore).
- A JS patch on the mail `Attachment` model marks office extensions as previewable (`isPdf`) and routes `defaultSource` through the conversion endpoint wrapped in the native PDF.js viewer. The *Download* button still returns the original file.
- Endpoint enforces: record read ACL, binary-field allow-list, extension allow-list, 50 MB size cap, 60 s conversion timeout. Returns 503 gracefully when LibreOffice is not installed.
- No new Python dependencies (stdlib `subprocess`/`tempfile` only).

### Test plan

- 14 unit tests covering the conversion helper and every HTTP branch of the endpoint (LibreOffice is mocked, so tests pass on runners without it).
- Manually validated on Odoo.sh (LibreOffice 7.3.7): xlsx/docx/pptx attachments preview correctly in the chatter.


alan196:
Update: the preview button is now hidden when LibreOffice is not available on the server, instead of opening the viewer with an HTTP 503 error (the availability is exposed through `session_info`). Failed conversions now also log LibreOffice's stderr.

Note for reviewers: the runboat image (oca-ci) does not ship LibreOffice, so on runboat the preview button will simply not appear for office files — the feature itself cannot be demoed there. To test functionally, run locally with `libreoffice` in the PATH (`apt-get install libreoffice`), attach an `.xlsx`/`.docx` to any record with a chatter and click the attachment.

alan196:
Improved spreadsheet rendering: XLSX/XLS/ODS are now converted with LibreOffice's `SinglePageSheets` export option (`calc_pdf_Export`), so each sheet becomes one full-size PDF page instead of being cut into print-size pages — columns are no longer split and the layout matches the original sheet much more closely. The conversion cache key now includes the file extension since the export filter depends on it.

alan196:
Added a loading indicator for slow conversions: since the PDF.js iframe gives no feedback while the server converts a large document, the FileViewer now shows a spinner ("Generating preview… Large files can take several seconds.") and warms the conversion cache with a fetch before mounting the iframe, which then loads instantly from the cache. Includes Spanish translation.

alan196:
Compatibility fix for LibreOffice < 7.4: the CLI JSON filter-options syntax (`pdf:calc_pdf_Export:{...}`) is silently ignored by older LibreOffice (verified on 7.3.7 — spreadsheets came out paginated again). The `SinglePageSheets` option is now seeded into the isolated profile's `registrymodifications.xcu` instead, which works from LO 7.2 onward (verified on 7.3.7 and 25.8). The cache key gained a version component to invalidate PDFs converted before this fix.

alan196:
Applying the same criteria as in #636 (thanks @pedrobaeza): this module is not a derivative of `attachment_preview` either, and it fits better in OCA/mail. Moved and renamed to `mail_attachment_office_preview` in OCA/mail#258. Closing this one in favor of that PR.

PR #636, opened 2026-08-24
[17.0][ADD] attachment_preview_xml: preview XML attachments as a collapsible tree

alan196:
## Summary

New module that renders XML attachments in the standard file viewer as an indented, collapsible tree instead of a wall of raw text.

### Context

`ir.attachment._check_contents` neutralizes xml-like mimetypes to `text/plain` for every user without write access on views, so XML attachments end up in the plain-text branch of the `FileViewer`. Machine-generated XML (electronic invoices such as CFDI or UBL, bank statements, EDI messages) is usually serialized on a single line, which makes that preview unreadable.

### How it works

- A JS patch on the mail `Attachment` model adds `isXml` (by extension, or by mimetype when it survives), removes those files from the `isText` branch and keeps them viewable.
- `XmlViewer` fetches the attachment, parses it with `DOMParser` and renders the tree with plain `<details>`/`<summary>` elements, so collapsing is native HTML: no library, no server-side conversion, no external service.
- Malformed XML, or content over 5 MB, falls back to the raw text instead of showing nothing.
- No Python code and no new dependencies: the module ships assets only.

### Test plan

- QUnit tests on the parser (hierarchy, attributes, comments, text-only nodes, malformed input), run in CI through an `HttpCase` that loads the module suite in a real browser.
- Manually validated on 17.0 with a single-line CFDI 4.0 attached to a chatter: the tree renders and every branch collapses; a deliberately malformed XML falls back to raw text.


pedrobaeza (approved):
Tested on runboat

HectorMerazChavez (approved):

pedrobaeza:
One question about the module name: with this name, it seems this is somehow a derivative of `attachment_preview` module, while it doesn't. What about naming it `mail_atttachment_xml_preview` and maybe put it on OCA/mail, as it's more related?

abr4hamw (approved):

alan196:
Thanks @pedrobaeza, you're right on both points: it is not related to `attachment_preview`, and mail is the natural place for it. Moved and renamed to `mail_attachment_xml_preview` in OCA/mail#257. Closing this one in favor of that PR.

PR #641, opened 2026-09-08
[19.0][MIG] document_page_project_task: Migration to 19.0

FHubert:
Migration of `document_page_project_task` to 19.0.

The module needed no code changes. Its dependency `document_page_project` migrated to 19.0 with a manifest-only diff, and all four inherited external IDs still resolve on 19.0 — `document_page.view_wiki_form`, `document_page.view_wiki_tree`, `document_page.view_wiki_filter` and `project.view_task_form2`, including the `button_box` div the stat button targets. The only change in this PR is the manifest version.

**Tested on a real 19.0 instance** (nightly `19.0-20260817`), not only through CI:

* installs cleanly, both views load, the m2m relation and the computed count are created;
* the `Wiki Pages` stat button appears on the task form and opens the filtered `Task Wiki` action;
* `task_ids` renders after `project_id` on the page form, with its project domain applied;
* `_check_task_project_consistency` correctly rejects a task belonging to another project;
* module tests on a clean database: `0 failed, 0 error(s) of 16 tests`.

`README.rst` and `static/description/index.html` are left to the post-merge readme bot: the pandoc available locally is older than the one CI uses, and regenerating with it introduced spurious list reformatting across the whole file rather than just the branch links.

One note for maintainers: this module is not listed in the 19.0 migration issue #559, since it was added to 18.0 after that issue was generated.


FHubert:
Updated: `pylint_odoo` flagged `W8161 prefer-env-translation`, so the module-level `_()` import is replaced with `self.env._()` in `_check_task_project_consistency`. Re-verified on the 19.0 instance afterwards — the constraint still raises with the interpolated message intact.

The migration commit is now three lines across two files: the manifest version and that translation call.

BhaveshHeliconia (approved):
Code review LGTM!

